PR #2, opened 2016-01-25
Fix crash on IE11 when previouslyFocused is null

gravityrail:
I was seeing this happen on IE11 on various versions of Windows.


davidtheclark:
Thanks, @gravityrail! I'll get a release out soon.

gravityrail:
:+1:

davidtheclark:
Published in 1.0.1.

PR #6, opened 2016-02-22
add support for nonModal option; fixes #4

### LINKED ISSUE #4: Non-modal mode
Right now when a focus-trap open it's essentially in "modal" mode, meaning that all interaction with the outside application is block.

For building non-modal dialogs, it would be nice to have a "non-modal" mode. Basically it could be a standard focus-trap but clicking outside of it would deactivate it.

heylookltsme:
This adds a `nonModal` option to the config object, allowing the focus trap to be deactivated on clicks outside of the trap element ([issue #4](https://github.com/davidtheclark/focus-trap/issues/4)).

### DIFF
--- a/index.js
+++ b/index.js
@@ -65,6 +65,12 @@ function deactivate() {
 
 function checkClick(e) {
   if (trap.contains(e.target)) return;
+
+  if (config.nonModal) {
+    deactivate();
+    return;
+  }
+
   e.preventDefault();
   e.stopImmediatePropagation();
 }


davidtheclark:
Thanks for the contribution.

One question: do you think we can stop at this simple solution, or do you think we need to also say that if you click back inside a deactivated nonModal trap, it activates again? Or if you tab through things and focus enters the trap again, the trap reactivates?

heylookltsme:
Hmm I was imagining that any re-activitation logic would live outside of focus-trap proper and be implemented by the person using it. Whatever originally triggers the focus trap can trigger it again if it needs to.

davidtheclark:
Yeah, I had the same thought. However, if there is *a way that non-modal focus-traps are always supposed to behave*, I think it'd be best to include that in the library rather than expect users to implement it. So here's my proposal for how the non-modal focus-traps might be expected to behave:

- Click outside and the trap deactivates, returning focus to the previously focused element or to whatever got clicked on, if what got clicked on receives focus.
- Once focus is outside the non-modal focus-trap, it's in a kind of semi-active state. In this state:
  - If you click inside the focus trap, it reactivates. (This means that the trap itself should have a tabindex of -1.)
  - If you tab through the page's focusable elements and end up re-entering the focus trap, the trap reactivates.
- If you close the focus trap by hitting Escape or doing something else that causes an actual `deactivate()` call, you end the semi-active state.

Does that seem too prescriptive, you think?

codrin-iftimie:
Hi, I'm trying to use the focus-trap-react to create a focus trap for an dropdown. Upon opening the dropdown I add a event listener to the document and check if we click outside of it to close it. The problem is that the focus trap blocks this event.

I suggest adding an option like: `blockEvents` that may have the default `["click", "focus", "keydown"]` instead a `nonModal` one.

What do you think?
cc: @heylookltsme  @davidtheclark

davidtheclark:
@codrin-iftimie Interesting idea. I was just working on an option to turn off the Escape key deactivation. And I see the value in making a click outside close the trap and then do it's regular thing, so I'll add that option. 

However, I don't see the value in being able to block other events. The point of the focus trap is to control focus, so I don't think you should be able to turn off the focus- or tab-controlling events. Right?

davidtheclark:
See 26f77d0661600024e836051a60996ae8714023c9 and 5f454dc111d083ca60e3b8d2cd455b1118857cac

davidtheclark:
I think that those two new features would address all of our real use-cases. @heylookltsme  and @codrin-iftimie do you agree? If so, I'll cut a new release.

codrin-iftimie:
Yes, thank you. I've set up a small demo here http://www.webpackbin.com/NyMGBuvTe. Works like a charm. There are two small issues: If you click outside the dropdown on the `Focusable`link if focuses the wrong element. I don't think of is as much of feature when it comes to clicking outside. Makes sense when pressing escape, but that is all.

A second comment: the activate function is rather restrictive with throwing errors when no tabbable element is found. I understand that accessibility wise there should be an element focussed when this happens, but maybe you should not activate at all when this happens. Throwing an error blocks the UI when used with React.

davidtheclark:
@codrin-iftimie: In 1.1.1 I tried to improve the `clickOutsideDeactivates` functionality, and I think it should fix this issue you brought up.

> A second comment: the activate function is rather restrictive with throwing errors when no tabbable element is found

I don't understand why you would want to activate a focus trap without any focusable elements. That seems like a mistake, and an error helps you catch that mistake. Do you have a use-case where it is desirable not to have a focusable element in the focus trap?

PR #12, opened 2016-07-12
Added an option for delaying the initial focus

nim23:
I have currently ran into a situation where if there is an animation in progress, voiceover on iOS acts pretty flaky and manages to focus on an focusable element at times, and not at times. Letting the animation finish and then proceeding with the focus seems to solve the problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Returns a new focus trap on `element`.
 - **onActivate** {function}: A function that will be called when the focus trap activates.
 - **onDeactivate** {function}: A function that will be called when the focus trap deactivates,
 - **initialFocus** {element|string}: By default, when a focus trap is activated the first element in the focus trap's tab order will receive focus. With this option you can specify a different element to receive that initial focus. Can be a DOM node or a selector string (which will be passed to `document.querySelector()` to find the DOM node).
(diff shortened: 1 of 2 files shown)


davidtheclark:
Why don't you add your delay before creating the focus trap?

nim23:
I was using the `focus-trap-react` component and that did not have any option to create a delay, so thought may be adding it here and passing it in as config option on the `component` would be the ideal solution. But in the end, I ended up creating my own React component to create the delay before activating the focus trap, which seems to work well for me. So please free to close this merge request, as the consumer of the `focus-trap` would have the ability to delay the `focus-trap` just like I did. Would it be sensible to create a merge request on the `focus-trap-react` component repo so that there is a option to delay the `focus-trap` because that is bit abstracted away in that component. And as I have seen with voiceover activating the focus-trap before the animation happens seems cause a quite a few problems.

davidtheclark:
> Would it be sensible to create a merge request on the focus-trap-react component repo so that there is a option to delay the focus-trap because that is bit abstracted away in that component.

Yeah, I think that's the ticket. Sounds like the problem is specific to that library. Thanks for the details!

PR #13, opened 2016-09-07
Do not throw Error if no focusable Element is available

swissmanu:
This change prevents `focus-trap` from failing because of a missing focusable element within its target element.

### Use Case
This change comes handy when using `focus-trap-react`: I had the specific situation where i wrap the `FocusTrap` around my generic modal dialog.
The dialogs content itself does not have focusable elements right from the start: First, there is only a loading text. The "real", focusable, content follows a split second later.

### Risks?
Digging into the code of `focus-trap` revealed that `tryFocus` was already robust enough to handle a `null`-return from `firstFocusNode`. Replacing line `115` with a simple `return null;` should not be dangerous at all then.

### Future Ideas
There could be a warning instead of a more strict error in the future. What kind of mechanism would you prefer to turn it off in production mode?

Suggestion: `process.env.NODE_ENV`?

### Kudos
Thank you for your module! `focus-trap` and `focus-trap-react` saved me a lot of implementation headache ☺️

### DIFF
--- a/index.js
+++ b/index.js
@@ -112,7 +112,7 @@ function focusTrap(element, userOptions) {
     if (!config.initialFocus) {
       node = tabbableNodes[0];
       if (!node) {
-        throw new Error('You can\'t have a focus-trap without at least one focusable element');
+        return null;
       }
       return node;
     }


davidtheclark:
@swissmanu Thanks! I'm not sure about this change, though. It makes sense in my mind that a focus trap cannot be opened until there is something to focus, even if it's the whole container. Otherwise focus already exists *outside* the trap, which is contrary to the purpose of the trap.

swissmanu:
thank you for your reply.
i totally see your point. there is no need to fail if nothing is focusable on the other hand.
personally, i would welcome a more fault tolerant behaviour. but after all, this is a decision you have to make for your library :)

davidtheclark:
@swissmanu I'm still thinking it makes sense to throw the error. It's like if you use `fs.readFile` and the file does not exist: you'll want to be informed that what you thought would happen is not actually going to happen. Because this might often be an error for people, I'd rather throw the error; and then if you or others want it silenced, you can use a try-catch to ignore it.

PR #14, opened 2016-09-07
Fix Outside Detection

swissmanu:
This change fixes the functionality which detects pointer events
outside of the focus trap.

Before, ANY pointer event, even inside of the focus trap, deactivated
it. By using the containers contain function with the events target,
this problem was fixed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -130,7 +130,7 @@ function focusTrap(element, userOptions) {
   // This needs to be done on mousedown and touchstart instead of click
   // so that it precedes the focus event
   function checkPointerDown(e) {
-    if (config.clickOutsideDeactivates) {
+    if (config.clickOutsideDeactivates && !container.contains(e.target)) {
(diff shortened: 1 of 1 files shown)


davidtheclark:
Thanks @swissmanu!

I'm not seeing the behavior you describe when I try out the demos on my phone (iPhone). Is this a bug that you've run into?

I do not remember off the top of my head why there's `checkPointerDown` in addition to `checkClick` — should `checkPointDown` just be deleted, replaced in event listeners with `checkClick`? I will have to think this through after I return from vacation in a couple of weeks. Sorry for the delay.

swissmanu:
never mind! enjoy your vacation :)

i did not test on any touch device. had the problems on a simple desktop browser.

davidtheclark:
@swissmanu I saw the problem in demo 3. Looks like this change fixes it. Thanks!

PR #16, opened 2016-09-12
Don't attempt to call 'blur' on 'document' (fixes #15)

### LINKED ISSUE #15: focus-trap throws errors on switching tabs in Firefox
**Steps to reproduce:**

1. Clone the repo and run `npm install`.
2. Build the demo bundle (`npm run demo-bundle`).
3. Open the demo page in Firefox (I used v48.0.1), then open another tab. Go back to the demo page's tab and open the browser console (so that you can see the error mentioned below).
4. Activate a focus trap (I chose to activate trap 1, but it doesn't make a difference).
5. Switch to the other tab in Firefox, then switch back to the tab with the demo page.
6. Observe that the following error is thrown in the console: `TypeError: e.target.blur is not a function`.

**What should happen:**

The error thrown in step 6 should not be thrown. Everything should just keep working as usual.

**Why it happens:**

Despite the fact that the spec disallows this, Firefox [fires a `focus` event on the `document` object itself](https://bugzilla.mozilla.org/show_bug.cgi?id=1228802), and has done so for quite some time now. However, the `document` object does not have a `blur` method. Since focus-trap adds a `focus` listener to the `document` object, the result is that, in Firefox, the `checkFocus` method receives the `document` object as `e.target`, and attempts to call `e.target.blur`, which is not a function.

**Ways to fix it:**

Here are a couple of options for fixing this (I assume it's something that focus-trap should at least work around for now, since Firefox has had this issue for a while and is unlikely to be fixed quickly). I'd like to submit a PR for either one of these, but figured it would also be worthwhile to ask which one is desired. I've locally written up both of them and checked (albeit quickly) that each one works. I'll include a PR for the first solution, below, as a proof-of-concept (which you can go ahead and merge in if it's ultimately the way you'd like to go); I'll be happy to modify it if necessary.

1. Check if `e.target` is equal to `document` in `checkFocus` and return early if so, or
2. Add listeners -- or at least the `focus` listener -- to `document.body` instead of `document`, or
3. Do both (1) and (2).

Any thoughts?

Thanks!

jmrog:
Fixes issue described in #15.

### DIFF
--- a/index.js
+++ b/index.js
@@ -143,10 +143,11 @@ function focusTrap(element, userOptions) {
   }
 
   function checkFocus(e) {
-    if (container.contains(e.target) || e.target === document) return;
+    if (container.contains(e.target)) return;
     e.preventDefault();
     e.stopImmediatePropagation();
-    e.target.blur();
+    // Checking for a blur method here resolves a Firefox issue (#15)
+    if (typeof e.target.blur === 'function') e.target.blur();
   }
 
   function checkKey(e) {
--- a/index.js
+++ b/index.js
@@ -143,7 +143,7 @@ function focusTrap(element, userOptions) {
   }
 
   function checkFocus(e) {
-    if (container.contains(e.target)) return;
+    if (container.contains(e.target) || e.target === document) return;
     e.preventDefault();
     e.stopImmediatePropagation();
     e.target.blur();


jmrog:
@davidtheclark Is there anything I can do to help this along?

davidtheclark:
Hi @jmrog. Thanks for the excellent issue and follow-up PR. I was out on vacation, so sorry I didn't respond sooner.

This seems like a non-intrusive fix to the issue you documented — which is great. However, I might prefer another fix, at least to start with. Since the problem isn't fundamentally that the target is the document, but rather that the document does not have a `blur` method, how about just checking if there is a `blur` method before calling it?

```js
if (e.target.blur) e.target.blur()
```

Will that fix the issue as well? Seems a little more targeted to me.

jmrog:
@davidtheclark No problem about response time; I figured something like that was the case!

Your suggestion sounds right-on to me, both for the reason you mentioned and also because it's probably more future-proof and defensive against other potential browser quirks. It does change the logic slightly, since `preventDefault` and `stopImmediatePropagation` will be called on the focus event fired in Firefox with `document` as its target. I'm almost certain that this won't actually change any behavior (there isn't any further for the event to propagate anyway, and `preventDefault` shouldn't actually have any effect on `focus` AFAIK), but I'd like to double-check that. After I do, and assuming all is well, I'll update the PR.

jmrog:
Didn't see any issues after making the change (and fix still works), so updated the PR. I went ahead and made the check ensure that `blur` is a function rather than just checking that it's truthy, just to be strict (if you'd prefer just the truthy check, though, I could go with that).

davidtheclark:
Thanks @jmrog!

Issue #18, opened 2016-09-30
About case with absence of focusable elements


vonagam:
About pull request #13.

There was stated that "I'd rather throw the error; and then if you or others want it silenced, you can use a try-catch to ignore it".

My use case: a popup which may or may not contain focusable elements.
If there is no such element and i use try-catch, then new focusTrap and its event bindings will not be created. User still will have ability to navigate and activate stuff under popup by tab or other means.

I propose to add some option to createOptions, which will determine, to throw an error or to return null from firstFocusNode if there is no focusable element.

davidtheclark:
@vonagam You should make the popup itself focusable. Focus needs to go somewhere.

vonagam:
There is possibility that there will be a focusable elements in popup's content.
If so, i want this element to gain focus.
So i can't use initialFocus option.

davidtheclark:
@vonagam Sorry but I still don't see the validity in not moving focus into the focus trap. Focus needs to enter the focus trap, or it's not a focus trap. There needs to be a focusable element, or it's not a focus trap. If you don't want to add a focusable element every time, whatever logic you use to conditionally insert the focusable element could also determine whether you use a focus trap.

vonagam:
It is focusable trap in a meaning that it does not allow any thing outside of a trap to be focusable.

I use it to write generic popup library and content is passed as argument. I have no way to know about what logic specific site uses to generate it.

A possible solution is to use "tabbable" package, since focus-group depends on it, to see if there is a focusable element, if no, than to set initialFocus to popup itself.

But it is preferably in my view to have specific option in "focus-group", since its dependencies can change and not part of api.

vonagam:
Maybe "fallbackFocus". If this option is present and there is no focusable element in container, than use this one, instead of throwing error.

Do you think that using "tabbable" directly is more desirable than adding new option?

davidtheclark:
`fallbackFocus` is a good idea. You should be able to do that fallback yourself using try-catch: if an error is caught, create a new focusTrap with the fallback focus. Right?

This has come up before, though, so if you're interested in a PR implementing a `fallbackFocus` option, that would be great.

vonagam:
Hm. Yeah, try-catch with setting initialFocus to a popup on catch is a solution that doesn't depend on "tabbable". But it makes code kinda ugly, though.

Ok, i'll try to create pull with `fallbackFocus`.

vonagam:
This issue can be closed now, as pull request was merged.

PR #19, opened 2016-09-30
Add fallbackFocus option

vonagam:
Added fallbackFocus option as proposed in #18.

Left description in README empty so that you can propose what to write there.

(Also fixed unneeded space before `returnFocusOnDeactivate` option)

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,9 +57,10 @@ Returns a new focus trap on `element`.
 - **onActivate** {function}: A function that will be called when the focus trap activates.
 - **onDeactivate** {function}: A function that will be called when the focus trap deactivates,
 - **initialFocus** {element|string}: By default, when a focus trap is activated the first element in the focus trap's tab order will receive focus. With this option you can specify a different element to receive that initial focus. Can be a DOM node or a selector string (which will be passed to `document.querySelector()` to find the DOM node).
+- **fallbackFocus** {element|string}: By default, an error will be thrown if the focus trap contains no elements in its tab order. With this option you can specify a fallback element to programmatically receive focus if no other tabbable elements are found. For example, you may want a popover's `<div>` to receive focus if the popover's content includes no tabbable elements. *Make sure the fallback element has a negative `tabindex` so it can be programmatically focused.* The option value can be a DOM node or a selector string (which will be passed to `document.querySelector()` to find the DOM node).
 - **escapeDeactivates** {boolean}: Default: `true`. If `false`, the `Escape` key will not trigger deactivation of the focus trap. This can be useful if you want to force the user to make a decision instead of allowing an easy way out.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -57,9 +57,10 @@ Returns a new focus trap on `element`.
 - **onActivate** {function}: A function that will be called when the focus trap activates.
 - **onDeactivate** {function}: A function that will be called when the focus trap deactivates,
 - **initialFocus** {element|string}: By default, when a focus trap is activated the first element in the focus trap's tab order will receive focus. With this option you can specify a different element to receive that initial focus. Can be a DOM node or a selector string (which will be passed to `document.querySelector()` to find the DOM node).
+- **fallbackFocus** {element|string}:
```
davidtheclark: Did you intend to add the rest of the documentation?

vonagam: "Left description in README empty so that you can propose what to write there."

I think it will be better if you will write description as you like.
You can write it here and i will amend it to the pull.

davidtheclark: 👍  How about this:

> By default, an error will be thrown if the focus trap contains no elements in its tab order. With this option you can specify a fallback element to programmatically receive focus if no other tabbable elements are found. For example, you may want a popover's `<div>` to receive focus if the popover's content includes no tabbable elements. *Make sure the fallback element has a negative `tabindex` so it can be programmatically focused.* The option value can be a DOM node or a selector string (which will be passed to `document.querySelector()` to find the DOM node).

vonagam:
Added seventh demo, where there are three buttons and fallbackFocus is set to trap itself.
First two buttons has tabindex="-1", last one is focusable, but initially hidden.
If you create trap while last button is hidden, then the trap gains focus.
If you create trap while last button is shown, then the button gains focus.

Left description for the demo empty. Will write what you will say.

Also did a little refactor of code change.

README.md
vonagam: Done

vonagam:
@davidtheclark any updates?

I added requested demo, but without description, as with README i left it to you to describe.

davidtheclark:
@vonagam Thanks for the reminder — somehow this slipped off my radar. Will look into it now.

demo/index.html
```diff
@@ -213,6 +213,32 @@ <h2 id="demo-six-heading">demo six</h2>
     </div>
   </div>
 
+  <h2 id="demo-seven-heading">demo seven</h2>
+  <p>
+  </p>
```
davidtheclark: Description can be 

> In this focus traps, the immediately visible buttons are not in the tab order. If you activate the trap in this state, the "fallbackFocus" option is used to focus the container. If, however, you make the focusable button visible, that will receive focus when you activate the trap.

davidtheclark:
Thanks @vonagam! Published in 2.1.0.

PR #27, opened 2017-03-17
Don't (re)focus and select if focus is already in the target element during trap activation

andrea-spotsoftware:
I have a focus trap for an input that is automatically activated when the user types something in that input.

During the activation the library is not checking if the node to focus is the same already focused, resulting in a misleading behavior: while user is inputing characters, the text become selected and continuing typing overwrites previously inserted chars. 

This PR just do this check in tryFocus() before focusing:

`
    if (node === document.activeElement)  return;
`

### DIFF
--- a/demo/js/index.js
+++ b/demo/js/index.js
@@ -5,3 +5,4 @@ require('./demo-four');
 require('./demo-five');
 require('./demo-six');
 require('./demo-seven');
+require('./demo-eight');
--- a/demo/index.html
+++ b/demo/index.html
@@ -11,7 +11,7 @@
 
 </head>
 <body>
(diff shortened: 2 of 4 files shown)


demo/js/index.js
```diff
@@ -5,3 +5,4 @@ require('./demo-four');
 require('./demo-five');
 require('./demo-six');
 require('./demo-seven');
+require('./demo-eight');
```
davidtheclark: Did you forget to commit this JS file?

andrea-spotsoftware:
I've added missing file, are you planning to merge this?

davidtheclark:
Makes sense, thanks!

PR #28, opened 2017-03-20
Trap callback

andrea-spotsoftware:
Ability to specify a callback function in the configuration options that will be called by active focus trap when prevents an event.


davidtheclark:
I am confused, @andrea-spotsoftware. I just merged the other PR. This, however, looks like the same thing with additional material. Can you please explain?

andrea-spotsoftware:
I worked on the previous branch, instead of creating a new branch from master! The only relevant commit is just the last one https://github.com/davidtheclark/focus-trap/pull/28/commits/4d0d72e77cd64779089df8360fe8f900d21c31e0 

Is the purpose of the PR clear?

davidtheclark:
> I worked on the previous branch, instead of creating a new branch from master!

Ok, please rebase for a clean PR.

> Is the purpose of the PR clear?

Actually not really. Could you please explain a use-case that motivates this change?

PR #32, opened 2017-04-18
Upgraded tabbable to latest version 1.0.6

iandevlin:


davidtheclark:
@iandevlin: Is there a reason we need to do this? Doesn't semver take care of the patch update?

iandevlin:
@davidtheclark To be honest, I don't know, does it? You would know more about it than I, so if you think it doesn't need to be done, then I believe you!

iandevlin:
It doesn't appear to work for me when I try to update `focus-trap`. The previous version pf `tabbable`is locked into this current release, so yes, I think that this PR is required.

davidtheclark:
@iandevlin: It's not required. If you delete `focus-trap` and reinstall it, the latest tabbable should come with it.

iandevlin:
@davidtheclark I did that, ran `yarn upgrade focus-trap` and it didn't work. It installed version 1.0.5 of `tabbable`.

iandevlin:
Hi, sorry to be a complete and annoying bore, but I would love this to be merged so I can update where I am using it so that it works in Edge and IE11.

davidtheclark:
@iandevlin: I'm pretty sure something is wrong with your package management. There is no reason that a new release with a minor patch should be necessary — this is exactly what semver is for.

davidtheclark:
Yep, just confirmed locally: a fresh install of this module (with either npm or yarn) also installs tabbable 1.0.6.

PR #33, opened 2017-04-20
Trap callback

andrea-spotsoftware:
Could be useful to know when focus trap prevents an event, because you could do some actions (popup) to inform user or something like this.
With this PR i've added the trapCallback function as optional parameter to the createFocusTrap options.

Demo updated.


davidtheclark:
@andrea-spotsoftware: Thanks for the PR! 

Looks like this is currently calling the callback whenever focus-trap responds to an event at all. For example, inside the focus-trap in the demo, the alert shows every time I hit tab. Is this your intention?

> you could do some actions (popup) to inform user or something like this.

Could you please explain more about the use-case? I don't feel comfortable making API changes until I fully understand the real-life use for them.

Lastly, could you please rebase this PR and fix the conflicts!

andrea-spotsoftware:
Ok, i'll try to describe my use case: I have an angular application with some inputs, and the viewModel behind those inputs is automatically updated when they **lose focus** (need this specific behavior to improve performances). 
Some of those inputs have the `focus-trap` attached, to prevent inputing invalid values. 
The `focus-trap` is activated by default from viewModel when field get focus, and deactivated by viewModel when inputed value is valid, but the problem is that **my viewModel is not changing until input has focus**. 
So I need a callback function in my code that fires everytime the `focus-trap` is preventing a user event (like Tab) to let me update my viewModel and - if the input value is valid - deactivating the trap.

In addition to that, I think that it could be useful for some application to explicitly inform a user that his attempts to exiting from the trapped input are not working because a trap is active.

What do you think about it?

davidtheclark:
@andrea-spotsoftware: Based on your description I'm not sure that what you need is a focus trap. If I understand correctly, focus is not actually cycling through a set of elements, but is in fact always returning to this one input until it is valid. That sounds less like a focus trap, more like validation behavior. If you have a function that runs whenever the input blurs, validating the value, couldn't that function return focus to the input if the value is invalid?

davidtheclark:
Closing as stale.

PR #34, opened 2017-05-30
Preventing the Bubbling on outside-buttons

petrogad:
I hit an issue where clicking outside of a trapped dialog *should* close the dialog however when I clicked outside on a button that had a touchend, it would re-trigger and flicker the dialog back.

Preventing the event cleaned up the issue.

Cheers
-Pete


davidtheclark:
I don't want to `preventDefault` on clicks outside. Most of the time when you click something outside the focus trap, you want your click to behave normally. In this special case you describe, I think you'll need to handle it in the code for you button.

PR #38, opened 2017-10-20
Feature/tabindex-1

MrEscaGreenwing:
Hi @davidtheclark :) 
I tried to make a minimum of changes to your code. I let the Browser handle the focus while tabbing. Afterwards `checkFocus()` checks if the focus left the trap and sets it back to the first or last tabbable element if necessary. Therefore I previously cache the tab-Event in a new object and read out if the user hit `Shift+Tab` or not. Depending on that information the focus will be set to the according element.
Tell me what you think about it. I'm open to suggested improvements.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -266,6 +266,45 @@
     </p>
   </div>
 
+  <h2 id="demo-nine-heading">demo nine</h2>
+  <p>
+    Initial focus on a element with a tabindex of -1 and no focus outline;
+    so when you "tab" you will get to the next or previous element according to the browser.
+    While tabbing through, the focus does not return to the initial focused element.
+  </p>
+  <p>
+    <button id="activate-nine">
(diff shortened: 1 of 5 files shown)


davidtheclark:
@MrEscaGreenwing `npm run start`

MrEscaGreenwing:
Well, I guess that was just too obvious... Thank you^^

Unfortunately you were right with demo three. That's why I returned to your original solution of manually moving the focus. Otherwise I couldn't track the correct order when leaving the trap because of other tabindexes. But I made an exception for elements with a tabindex of -1 and adapted my new solution to them. I went through all the demos and it seems to work as expected. Furthermore I added a new demo 9 for my use case.

davidtheclark:
Thanks again, @MrEscaGreenwing. Just published a new version that includes these changes.

PR #40, opened 2017-11-29
Support Web Components

DRiFTy17:
Added support for traversing through shadow roots when determining the active element. Adjusted how the event target is determined to get it from event.path/composedPath where supported, and fall back to a custom composed path implementation where not supported.

Let me know if you see anything that may not support your intended browsers. 

For focus-trap to work with web components it will also require the the PR I submitted to tabbable for web component support so this should only get merged once the one from tabbable is merged, otherwise the tabbable elements will never get picked up correctly.


davidtheclark:
Closing: see https://github.com/davidtheclark/focus-trap/issues/39#issuecomment-408714143

PR #41, opened 2017-12-05
Add TypeScript type declarations

TedDriggs:
Added per instructions on TS [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/by-example.html).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,83 @@
+declare module "focus-trap" {
+  /**
+   * A DOM node, a selector string (which will be passed to
+   * `document.querySelector()` to find the DOM node), or a function that
+   * returns a DOM node.
+   */
+  export type FocusTarget = HTMLElement | string | { (): HTMLElement };
+
+  export interface Options {
+    /**
+     * A function that will be called when the focus trap activates.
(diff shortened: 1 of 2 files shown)


davidtheclark:
:wave: @TedDriggs. Since I don't use TypeScript myself, I'm not really interested in maintaining the TS declaration files. This is something that TS has solved in other ways, right?

TedDriggs:
@davidtheclark There is the option to register them with the DefinitelyTyped repository, but the consuming package has to know that those types are available from the external type repo.

Are you planning to change the API for the package? If not, then these shouldn't require maintenance. If you are, I'm happy to take on owning any needed updates to the declaration files when you update the docs (just @ me or assign a ticket). Our team is taking a dependency on focus-trap-react so we're motivated to have these stay current.

davidtheclark:
Ok, @TedDriggs — I'll give this a chance. I'll just leave it to TypeScript users to update as needed.

davidtheclark:
@TedDriggs I really made a mess trying to release this morning ... must be groggy ... but I'm pretty sure I got it out in v2.4.2.

Issue #42, opened 2018-01-05
Allow to interact on some outside elements


gejgalis:
I'm having an issue using focus-trap with KendoUI and pickers components like Date Picker. When Kendo Date Picker is placed inside trapped container, then I want to open the picker and click inside picker to select a day. focus-trap prevents clicks inside Kendo picker, beacuse Kendo reparents its pickers directly under `<body>` element, so pickers are treated as outside elements by focus-trap.

To solve this problem I'm going to create PR with additional configuration option:

`includeOutsideElement: Function(element:HTMLElement):boolean ` a function which will be called when focus-trap detects click/mousedown/touchdown events outside trapped container. `true` result of this function will not prevents and not stops above events. Developer needs to prevent those events on his own handlers. Kendo prevents mousedown events inside pickers, so focus won't be lost.

davidtheclark:
Hello @gejgalis. I have a question: When a DatePicker opens, shouldn't focus be able to enter that picker? If `includeOutsideElement` only *allows clicks* in the outside element, you can't get to the outside element by tabbing, right?

If you agree that we may want to allow focus to move between the elements (the original form and the displaced DatePicker), maybe the solution is to allow focus trap to accept multiple elements?

gejgalis:
Hello @davidtheclark . Kendo pickers prevents mousedown event, so you can't jump with focus from date input to date picker - you can only use a mouse to choose a day, so focus always stays in date input.  [See date picker demo](https://demos.telerik.com/kendo-ui/datepicker/index)

Allowing multiple elements upfront as one logical trap container is not always possible (using HTMLElement reference), because some of those elements are created dynamically after click. It would be only possible by configuring string selector (not HTMLElement reference). If you are OK with that - it will work for me too.

gejgalis:
@davidtheclark you want me to prepare PR for multiple targets?

davidtheclark:
@gejgalis Sorry for the delayed answer here. If you're still interested in trying to add this feature, I'm open to a PR.

davidtheclark:
This is certainly an edge case and nobody has yet finished a PR with a workable API. So I'm going to close this issue for now. If you are invested in this use case and want to work out the details with me, please open a new issue.

PR #43, opened 2018-01-05
Allow to interact on some outside elements #42

gejgalis:
Please let me know if it is acceptable. Maybe configuration name should be different?

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Returns a new focus trap on `element`.
 - **escapeDeactivates** {boolean}: Default: `true`. If `false`, the `Escape` key will not trigger deactivation of the focus trap. This can be useful if you want to force the user to make a decision instead of allowing an easy way out.
 - **clickOutsideDeactivates** {boolean}: Default: `false`. If `true`, a click outside the focus trap will deactivate the focus trap and allow the click event to do its thing.
 - **returnFocusOnDeactivate** {boolean}: Default: `true`. If `false`, when the trap is deactivated, focus will *not* return to the element that had focus before activation.
+- **includeOutsideElement** {Function(HTMLElement):boolean}: Default: Function returning `false`. A Function called when focus-trap detects interaction outside trapped container (mousedown, click, touchstart). `true` result will treat this element as if it was inside trapped container. This option may be useful for pickers components where part of UI is attached directly under `<body>` element, because focus-trap prevents interaction outside trapped container by default. WARNING: click/mousedown/touchstart events must be prevented and stopped in your code to prevent losing focus from trapped container.  
 
 ### focusTrap.activate()
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -54,6 +54,17 @@ declare module "focus-trap" {
      * deactivate the focus trap and allow the click event to do its thing.
      */
     clickOutsideDeactivates?: boolean;
+
+    /**
+     * Default: Function returning `false` A Function called when focus-trap detects interaction outside trapped
+     * container (mousedown, click, touchstart). `true` result will treat this element as if it was inside trapped
+     * container. This option may be useful for pickers components where part of UI is attached directly under `<body>`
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -61,6 +61,7 @@ Returns a new focus trap on `element`.
 - **escapeDeactivates** {boolean}: Default: `true`. If `false`, the `Escape` key will not trigger deactivation of the focus trap. This can be useful if you want to force the user to make a decision instead of allowing an easy way out.
 - **clickOutsideDeactivates** {boolean}: Default: `false`. If `true`, a click outside the focus trap will deactivate the focus trap and allow the click event to do its thing.
```
davidtheclark: @gejgalis: Thanks for your patience. I think I understand your goals here and agree that it's useful. I have one source of hesitation: I worry that the relationship of `clickOutsideDeactivates` and `includeOutsideElement` is kind of confusing. What do you think about changing this up a little so we do not add an `includeOutsideElement` option but instead allow `clickOutsideDeactivates` to be a function? These would be the rules for `clickOutsideDeactivates`:

- If `true`, all clicks outside the focus trap deactivate it.
- If `false`, no clicks outside the focus trap deactivate it.
- If a function, when there's a click outside the focus trap the clicked element is passed to the function and a boolean is returned. If the function returns `true`, the focus trap is deactivated. If the function returns `false, the focus trap is not.

I'm thinking the advantage is that we no longer have a conflict between two options: `clickOutsideDeactivates` is the only one you set to control what outside clicks do.

How does that sound?

gejgalis: I'm not sure if your proposal solves my problem. I don't want to activate/deactivate focus trap when clicked outside - I just want to skip preventing click events outside.

I think your previous proposal was better: istead of adding new option `includeOutsideElement` we could allow to setup multiple elements? What do you think @davidtheclark ?

davidtheclark: > I'm not sure if your proposal solves my problem.

@gejgalis I don't understand how it doesn't? — maybe the weirdness of the Kendo UI is confusing me. I thought the goal is to allow you to click on (but not focus) *some* external elements (the datepicker) without deactivating the focus trap? If so, I think it makes sense to say, essentially, "clickOutsideDeactivates when you click anywhere on the screen except within the datepicker" — and that's what a filter function could allow for.

gejgalis: @davidtheclark I'm convinced now. It makes sense and it should work for me. You want me to prepare PR?

davidtheclark: Glad it made sense to you. Yeah, a PR would be great, @gejgalis!

pgn-vole:
Can this PR be looked at again? We need the same feature, for React Components not kendo ones though. 
I think the initial implementaion makes more sense that your new proposed one @davidtheclark. One doesn't want to deactivate the focus trap entirely but simply let some node to not be ignored when the focus trap is active.

davidtheclark:
@pgn-vole You're welcome to open another PR following through on this plan. I don't think I understand your objection, though.

> the goal is to allow you to click on (but not focus) some external elements (the datepicker) without deactivating the focus trap

Is that not your goal?

pgn-vole:
@davidtheclark That's exactly my goal. 
Your suggestion, to re-use the API `clickOutsideDeactivates` does more than just this though. It entirely disable the focus trap when clicking on an outside element. We don't need that, we just would like for some clicks to pass through while keeping the focus trap enable.
Does that make sense?

davidtheclark:
> Your suggestion, to re-use the API clickOutsideDeactivates does more than just this though. It entirely disable the focus trap when clicking on an outside element.

No: my suggestion is to make `clickOutsideDeactivates` accept a function that you can use to determine whether a click on an outside element disables the focus trap or not.

pgn-vole:
I see, so they are actually two different use cases :)
I will work on another PR to solve the problem I'm facing then.

davidtheclark:
Closing: see https://github.com/davidtheclark/focus-trap/issues/42#issuecomment-408714255

PR #51, opened 2018-05-22
Add delay before autofocus

pgn-vole:
Add delay before autofocus when the focus trap get activated.
It should prevent any focusable elements within the focus trap to capture the event that triggered the activation of the focus trap.
See https://github.com/davidtheclark/focus-trap/issues/46 for more details.


davidtheclark:
Thanks, @pgn-vole! Looks good.

Before we merge, please 
- remove `.idea/workspace.xml` from the commit; 
- clarify in the demo description that *any* key press on the "reveal" button should open the focus trap (or could we make that more restricted and realistic?).

Then let's merge and release.

FabienDeshayes:
@davidtheclark any chance of seeing this merged soon? We're quite interested to use this :)

davidtheclark:
Thanks, @pgn-vole: looks good 👍 I'll release it in a second.

PR #53, opened 2018-05-23
isOutsideElementFocusable API

pgn-vole:
Add `isOutsideElementFocusable` API to allow elements outside of the container to be part of the tab cycle and be focusable.

It should fix https://github.com/davidtheclark/focus-trap/issues/42

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ Returns a new focus trap on `element`.
 - **escapeDeactivates** {boolean}: Default: `true`. If `false`, the `Escape` key will not trigger deactivation of the focus trap. This can be useful if you want to force the user to make a decision instead of allowing an easy way out.
 - **clickOutsideDeactivates** {boolean}: Default: `false`. If `true`, a click outside the focus trap will deactivate the focus trap and allow the click event to do its thing.
 - **returnFocusOnDeactivate** {boolean}: Default: `true`. If `false`, when the trap is deactivated, focus will *not* return to the element that had focus before activation.
+- **isOutsideElementFocusable** {function}: A function that returns if an element on the page should be part of the tab cycle. By default only descendant of the container are focusable. 
 
 ### focusTrap.activate([activateOptions])
 
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Returns a new focus trap on `element`.
 - **escapeDeactivates** {boolean}: Default: `true`. If `false`, the `Escape` key will not trigger deactivation of the focus trap. This can be useful if you want to force the user to make a decision instead of allowing an easy way out.
 - **clickOutsideDeactivates** {boolean}: Default: `false`. If `true`, a click outside the focus trap will deactivate the focus trap and allow the click event to do its thing.
 - **returnFocusOnDeactivate** {boolean}: Default: `true`. If `false`, when the trap is deactivated, focus will *not* return to the element that had focus before activation.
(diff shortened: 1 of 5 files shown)


davidtheclark:
@pgn-vole: I don't think it makes sense to allow focus on an element *outside* of the focus trap. https://github.com/davidtheclark/focus-trap/pull/43 was about allowing *clicks* on outside elements, not focus.

It sounds like what you want is for a focus-trap to be able to contain nodes that are not all part of the same branch on the DOM tree, right? That way you'd also be able to tab across them. I think a solution to this would be to allow for the creation of focus-traps that span multiple elements.

pgn-vole:
@davidtheclark , I took this approach because while coding the clickable feature I thought that there was a way to make it even more generic by not limiting the scope of the API to "click" but allowing the elements to be "focusable". I also thought of the approach of passing supplemental DOM elements containers but because the kind of elements that this API targets are transient ones (modal, flyout ect) It seems that it would have been much handy from the consumer POV to to pass a function which is executed on the fly.

Anyway, I'm happy to update the PR to limit the scope to only care about the click events.

davidtheclark:
@pgn-vole Now we have two very similar options: `clickOutsideDeactivates` and `isOutsideElementClickable`. As I suggested here https://github.com/davidtheclark/focus-trap/pull/43#discussion_r165828158, how about merging those into a single option? Otherwise they are in danger of contradicting each other.

pgn-vole:
@davidtheclark they are not doing the same thing so they should not be merged.
Cf my comment on the the other thread, `clickOutsideDeactivates` deactivate the focus trap entirely. We don't want that. We just want clicks to pass through while keeping the focus trap activated.

davidtheclark:
`clickOutsideDeactivates` is a boolean option that indicates whether a click outside the focus trap should deactivate the trap or not. `isOutsideElementClickable` *is* doing the same thing: it's determining whether a click on some element outside the focus trap should deactivate the trap. It's just doing with a function instead of a boolean.

pgn-vole:
That's not exactly the same because clickOutsideDeactive actually Deactivate the focus trap. isOutsideElementClickable doesn't do that. It keeps the focus trap enable and allow click to go through, that's the point of it. The feature should allow to be able to click on elements that are not descendant of the "focus-trapped" element WHILE keeping the focus trap enable. 
Doesn't it make sense?

FabienDeshayes:
@pgn-vole @davidtheclark @gejgalis
I think both PRs do pretty much the same thing.
#43 adds an option `includeOutsideEelment` that is a function returning a Boolean for a given DOM element. If that function returns `true`, then the trap doesn't capture the click / mousedown / touchstart and it goes through.
In this current PR #53 the option is called `isOutsideElementClickable` that is also a function returning a Boolean for a given DOM element, but the difference is that it only works for clicks and not mousedown / touchstart.

I think #43 covers more use cases and should be merged.

Thoughts?

davidtheclark:
Hi @FabienDeshayes. I'm still pretty convinced that the solution to #42 is to make `clickOutsideDeactivates` a function that receives the element that was clicked and returns a boolean indicating whether that click means the track should deactivate or not. I do not see the need for a new option. 

Neither PR does that yet. If you'd like to open another PR doing that, go for it!

davidtheclark:
Closing: see https://github.com/davidtheclark/focus-trap/issues/42#issuecomment-408714255

PR #55, opened 2018-07-04
Set container as the default focus target

Dan503:
The current implementation of focus-trap has a major accessibility issue that can easily be resolved by making the focus-trap container the default element that receives focus when the focus trap is activated.

For example, have a look at demo 1 on the [demo site](http://davidtheclark.github.io/focus-trap/demo/).

This is what the UX is like for a screen reader user using [NVDA](https://www.nvaccess.org/download/) trying to use demo 1 running the current focus-trap default implementation:

1. Focus on activate button
2. _"activate trap 1 [button]"_
2. Press enter
3. _"with [link]"_

Now this is what it is like if you accept my pull request:

1. Focus on activate button
2. _"activate trap 1 [button]"_
2. Press enter
3. _"Here is a focus trap with [link] some [link] focusable [link] parts. Deactivate trap 1"_

See how much more accessible this default functionality is for screen reader users?

----

There is another bonus up side to accepting this pull request. The current version errors if there are no focus-able elements in the container. That restriction is no longer an issue. It will be a lot harder to make focus-trap error after this pull request. As long as the focus trap element isn't hidden and doesn't have it's tabindex property removed, it will not error.

This is a major breaking change in the API since it changes core default functionality. I think it is worth it though since it encourages better accessibility practices and enables the ability to use the plugin on elements with no tabbable links.

One thing, can you check that the delay demo is still working as expected? That demo confused the hell out of me :/


Dan503:
Oh, I just realised that I should add a demo to test what happens if there are no links in the focus-trap.

I don't have time to do it right now though.

davidtheclark:
https://github.com/davidtheclark/focus-trap/issues/56 is closed now.

Issue #58, opened 2018-07-15
Can we avoid hijacking tab?


davidtheclark:
Right now this library hijacks <kbd>Tab</kbd>. It relies on [tabbable](https://github.com/davidtheclark/tabbable) to determine the tabbable nodes within the focus-trap, and their order, then moves focus accordingly when you hit <kbd>Tab</kbd>, instead of letting the browser do its default thing.

There are a couple of issues with this approach:

- tabbable does not (and probably *cannot*) perfectly match the browser's default behavior. The main symptoms of this are that tabbable is not going to pick up on [iframes](https://github.com/davidtheclark/tabbable/issues/23) or [shadow DOM](https://github.com/davidtheclark/focus-trap/issues/39) — so in a focus-trap those elements won't receive focus; and it also [can't handle radio sets](https://github.com/davidtheclark/tabbable/issues/28).
- It seems unfortunate: it would be nice if we didn't have to hijack default browser behavior but could somehow just rein it in a little.

If we could find an approach that fixes those issues *and still works in all the ways focus-trap currently works* — that would be great. **I'm opening this issue to hear any ideas anyone has for solving this problem. Input welcome!**

---

I tried experimenting a bit with an alternative approach today. I used tabbable only to grab the first and last tabbable nodes in the trap. Instead of listening to <kbd>Tab</kbd> events, I listened to `focusout` events: if `event.relatedTarget` comes *before*  the previously focused node (so you're shift+tabbing), I'd focus the last tabbable node; if `event.relatedTarget` comes *after*, I'd focus the first tabbable node. I ran into 2 bad roadblocks:
- This seemed promising on a few of the demos, but then crashed and burned with demo 3. When `tabindex` is deliberately set anywhere on the page it messes up the whole approach.
  - Maybe this problem could be accepted as a caveat and we could recommend strongly against the use of `tabindex 1+` (which some accessibility experts also strongly discourage [Paciello Group](https://webaim.org/techniques/keyboard/tabindex#tabindex), [WAI-ARIA Practices](https://www.w3.org/TR/wai-aria-practices/#dialog_modal), [Google Web Fundamentals](https://developers.google.com/web/fundamentals/accessibility/focus/using-tabindex)).
- The first and last focusable items in the trap still need to be recognizable to tabbable.

davidtheclark:
Closed by #59.

PR #59, opened 2018-07-15
Overhaul for alternative focus management

davidtheclark:
See https://github.com/davidtheclark/focus-trap/issues/58: I'm trying to avoid hijacking Tab.

Two known caveats with this approach so far (see #58 for more details):
- You do need a regular focusable element before and after your iframe (or radio set or shadow DOM node, etc).
- You cannot use positive tabindexes. I'm not sure yet whether you can't use them *anywhere on the page* or just within the focus trap. 

Also did some refactoring and renamed all the demos, since I eliminated a couple. So this is a big overhaul.

- [x] Test in Firefox
- [x] Test in Safari
- [x] Test in IE11
- [x] Test in Edge
- [x] Pass on Travis


davidtheclark:
Well, with the caveats above this PR now seems to work well for all of the demos — which includes demos with iframes and radio sets. I'm wary of merging the change, though, because of how finicky this is, so how likely it is that I'm overlooking something. **If anybody wants to help push this change through, I'd appreciate help validating the accessibility of the change by checking this branch with screen readers and other browsers.**

PR #62, opened 2018-07-29
Return focus on outside click if clicked element is not focusable

### LINKED ISSUE #60: Make clickOutsideDeactivates=true return the focus if the click target is not focusable
I'm having an issue building a drawer component using a backdrop (shadow overlay). When the backdrop is clicked, the drawer should be closed. Setting `clickOutsideDeactivates` to `true` is required for the click event on the backdrop to be fired, but it seems that enabling this setting does not return the focus to the button used to activate the drawer.

This is understandable because the element clicked outside the focus trap might be a focusable element itself. Thats why I would suggest to add an extra config option to toggle if the focus should be returned, or do an extra check if the element outside is focusable.

davidtheclark:
Closes #60. Replaces https://github.com/davidtheclark/focus-trap/pull/61.

Thanks @egonolieux! I ended up adding `tabbable.isFocusable` to the Tabbable library to make this cleaner.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,8 +4,9 @@ var xtend = require('xtend');
 var listeningFocusTrap = null;
 
 function focusTrap(element, userOptions) {
+  var doc = document;
   var container =
-    typeof element === 'string' ? document.querySelector(element) : element;
+    typeof element === 'string' ? doc.querySelector(element) : element;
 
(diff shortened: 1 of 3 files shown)


egonolieux:
Cheers!

PR #69, opened 2018-11-29
Upgrade the version of tabbable to 3.1.1

abhiomkar:
This change allows `focus-trap` to be used in React components rendered on server side.

See [tabbable PR#33](https://github.com/davidtheclark/tabbable/pull/33)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6123,9 +6123,9 @@
       }
     },
     "tabbable": {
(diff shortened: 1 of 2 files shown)


davidtheclark:
Thanks @abhiomkar. I'll merge this. To be clear, though, you should be able to get the latest version of tabbable installed without needing this change.

PR #70, opened 2018-12-04
Pass triggering events for deactivating focus trap to callbacks

waiwaing:


waiwaing:
This change allows the callback to identify the triggering event. In my specific scenario, we want to allow a click outside of the focus trap to be captured if the target is one of a specific number of elements, and open a different focus trap accordingly. Without knowing the triggering event, what happens is:  
1. The click outside deactivates the focus trap
2. The callback closes the focus trap div
3. The click passes through and opens a new div which has the focus trap

If the event is known, the flow can be as follows:  
1. The click outside deactivates the focus trap
2. The callback inspcts the event target, replacing the focus trap div with a new div which has the focus trap

In the first workflow, there are two renders, causing a flash of the content behind the modal div, whereas in the second workflow, there is only one render.

Flydiverny:
@davidtheclark  Any updates on this?

This would help us as well, our use case is that we have a button opening a menu.
Clicking the button again to close it is not possible as the deactivation will trigger first so the button would just open the menu again.

davidtheclark:
I'm fine with this feature, but I have a few requests:

- Instead of just passing the event to `onDeactivate`, let's pass an object with an `event` property. That way this pattern is extensible without worrying about argument order.
- Please add a changelog entry.
- Please add documentation.
- Please add to or adjust the demos so they illustrate how this feature works.

jtomaszewski:
@waiwaing @Flydiverny would https://github.com/davidtheclark/focus-trap/pull/88 solve your problem?

> our use case is that we have a button opening a menu.
Clicking the button again to close it is not possible as the deactivation will trigger first so the button would just open the menu again.

I showed how such thing can be done in that PR - see https://www.dropbox.com/s/ryso8l3u5s14c23/focus-trap-allowoutsideclick-demo.mov?dl=0 .

stefcameron:
Since David's [requests](https://github.com/focus-trap/focus-trap/pull/70#issuecomment-493777322) weren't addressed on this old PR, I'm closing it.

PR #77, opened 2019-03-13
Remove the outdated yarn.lock file

stof:
The package-lock.json file is uptodate, so it looks like this package is intended to be worked on using npm rather than yarn.

WebStorm would use Yarn when a yarn.lock file is there, so removing it means WebStorm users would get messages about the right tool.


davidtheclark:
Thanks @stof.

PR #78, opened 2019-03-13
Optimize the usage of tabbable

### LINKED ISSUE #76: Activating a trap updates the tabbable nodes twice
In my experience, the bottleneck of activating a focus-trap is the tabbable library when the trap contains lots of tabbable nodes.
When activating a trap, the list of tabbable nodes is updated twice: one directly in `activate` and one inside `addListeners`. Would it be possible to combine those to update the list only once ?

stof:
Fixes #76 and also updates to the newest faster tabbable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 5.0.0
+
+- Update Tabbable to improve performance (see [Tabbable's changelog](https://github.com/davidtheclark/tabbable/blob/master/CHANGELOG.md)).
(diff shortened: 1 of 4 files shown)


stof:
I used 5.0.0 as version when writing the changelog, because you decided to make the tabbable refactoring a major version (and so making it a minor in focus-trap would be weird)

davidtheclark:
Thanks @stof. All of the demos work as expected for me 👍

stof:
@davidtheclark it would be great to tag a 5.0.0 release containing that optimization (maybe merging #79 first though).
I would like to send a PR to material-components-web to use the new optimized version.

davidtheclark:
Yeah, @stof, I was waiting to resolve #79. I'll try to get a release out within a week.

davidtheclark:
Just released 5.0.0 🐣

PR #79, opened 2019-03-13
Register listeners explicitly as not passive

stof:
Listeners are preventing the default action in some conditions. Recent Chrome versions have an intervention registering root-level listeners as passive by default for some events. So listeners have to be declared as non-passive explicitly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -143,10 +143,22 @@ function focusTrap(element, userOptions) {
       tryFocus(getInitialFocusNode());
     });
     doc.addEventListener('focusin', checkFocusIn, true);
-    doc.addEventListener('mousedown', checkPointerDown, true);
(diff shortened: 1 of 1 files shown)


stof:
@davidtheclark passing an object as third argument will be casted to a `true` boolean for older browsers. As that's the value we want, that's fine.

stof:
@davidtheclark do you still want me to perform some feature detection to pass an actual boolean for `capture` here ? That would increase the code size for no benefit. The current code works fine in older browser (we would need feature detection to register a listener in the bubbling phase, but these are in the capture phase)

davidtheclark:
No worries, @stof. I just wanted to confirm that this works in older browsers — that they don't error if the value is not a boolean. I just tested on IE11, and sounds like you've also tried this on older browsers; so I think we're good to merge & release.

PR #83, opened 2019-05-16
Clear delay before autofocus in deactivate method

SonyaOrlova:
Fixes https://github.com/davidtheclark/focus-trap/issues/81

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,8 @@
 var tabbable = require('tabbable');
 var xtend = require('xtend');
 
+var activeFocusDelay;
+
 var activeFocusTraps = (function() {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -90,6 +92,8 @@ function focusTrap(element, userOptions) {
   function deactivate(deactivateOptions) {
     if (!state.active) return;
 
+    deactivateDelay(activeFocusDelay);
```
davidtheclark: I think we could use `clearTimeout` directly here instead of a wrapper function.

SonyaOrlova: done)

Issue #30, opened 2017-03-31
Not compatible with ES2015 / ES6 (import)


sam-s4s:
Hmm, I would be great if you could use this package in ES2015 / ES6 land (ie. with Rollup).

Currently only seems to be able to be used with require(), which isn't compatible.

Needs to be able to be imported...

import x2js from 'x2js';
-or even-
import * as x2js from 'x2js';

Any chance of that happening? :)

Also typings? :D

davidtheclark:
This seems like a major problem with Rollup, right? Is there really no way to easily consume CommonJS modules?

sam-s4s:
Yeah, I can't get my head around it lol

It seems like they just don't care about anything that isn't an ES6 module... Makes me sad!

davidtheclark:
@sam-s4s: how about https://github.com/rollup/rollup-plugin-commonjs? Found that recommendation in the docs.

sam-s4s:
Yeah, using this already (for rxjs), however it still doesn't let you use require() :( Only import

davidtheclark:
@sam-s4s: With that plugin I think you should be able to `import` modules that use CommonJS. If that doesn't work for you, sounds like you might need to seek help in that repo, rather than here.

mkay581:
Hello, @davidtheclark I am running into this same issue which makes this package unusable for me unfortunately. I am using Lit without any build process. es6 imports/exports are now supported by browsers and it would help make this package consumable using polymer/lit if it supported es6 import/exports. I do this with my packages also to ensure interoperability.

patrickk:
Running into this problem as well @mkay581, I ended up modifying both tabbable and focus-trap to use `export default` and it seems to work, but are there any plans to migrate over to es6 syntax in the future? @davidtheclark

mkay581:
Yeah I'd strongly encourage supporting es6 imports/exports. More and more people are using them and they're growing to be a community-wide standard. It wouldn't take much to update this package to use it. @davidtheclark would you accept a PR to do this?

juampi:
Any updates on this @davidtheclark ?

davidtheclark:
If anybody would like to submit a PR with minimal changes that add a build script to address this — without any breaking changes for non-ES2015 users — I'd be open to taking a look.

gertcuykens:
I am stuck too on this example https://jsbin.com/saqoyinufi/edit?html,console,output
ref: https://github.com/material-components/material-components-web/issues/4507

[...]

PR #84, opened 2019-05-27
Add ES6 Support to focus-trap

brian-lagerman:
This PR should open up focus-trap for use in ES6, solving issue  #30

It:
- Adds rollup with 3 plugins to Dev dependencies
- Uses rollup-modify to turn index.js into valid ES6 syntax
- Bundles tabbable since it isn't a module
- Pushes the result to the dist folder
- Marks "dist/focus-trap.mjs" as the module file in package.json
- Sorts scripts alphabetically (sorry)

It does not:
- Minify the module file (could be done with either uglify-js or rollup-terser)
- Add any demo code/section for module loading

I did create a [CodePen](https://codepen.io/brian-lagerman/pen/OYEjOX) using the resulting file, and it appears to work as intended.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,0 +1,33 @@
+import commonjs from 'rollup-plugin-commonjs';
+import resolve from 'rollup-plugin-node-resolve';
+import modify from 'rollup-plugin-modify';
+
+export default [
+  {
+    input: 'index.js',
+    output: {
+      file: 'dist/focus-trap.mjs',
+      format: 'esm'
+    },
+    plugins: [
+      modify({
+        // swap tababble require load to ES6 import
+        "var tabbable = require('tabbable');":
+          "import tabbable from 'tabbable';",
+        // ES6 module support implicitly means Object.assign support, so xtend can drop out
+        "var xtend = require('xtend');": '',
+        'xtend(': 'Object.assign(',
+        // swap focus-trap from node module.exports to ES6 named export
+        'function focusTrap(': 'export function focusTrap(',
+        'module.exports = focusTrap;': ''
+      }),
+      resolve(),
+      commonjs({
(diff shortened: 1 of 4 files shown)


rollup.config.js
```diff
@@ -0,0 +1,33 @@
...
+    plugins: [
+      modify({
+        // swap tababble require load to ES6 import
+        "var tabbable = require('tabbable');":
+          "import tabbable from 'tabbable';",
+        // ES6 module support implicitly means Object.assign support, so xtend can drop out
+        "var xtend = require('xtend');": '',
+        'xtend(': 'Object.assign(',
+        // swap focus-trap from node module.exports to ES6 named export
+        'function focusTrap(': 'export function focusTrap(',
```
dougalg: Can browserify/rollup not handle multiple targets? Having multiple build tools seems less than ideal? I know webpack could handle generating multiple outputs of various types if wanted.

To me, the ideal pattern would be:
1) Update the src to be es6
2) Build then just minifies source to a `dist/**.mjs` file for es6 support
3) Build also runs rollup or webpack or browserify to generate the legacy commonjs support file

For the `xtend` use case, I feel like that can just be left out and library users can specify (via polyfill) how to handle `Object.assign` uses.

stefcameron:
@brian-lagerman Thanks for this PR! I'm a new maintainer, trying to get things back on track here. 😄  I'm closing this in favor of #94 which adds ESM support with rollup in what seems to be a cleaner way.

brian-lagerman:
@stefcameron I couldn’t agree more. This PR was a very non-conventional attempt to provide ESM support while still adhering to the previous maintainers desire to not modify the original code base. Upgrading the source to es6 is obviously a much better approach.

PR #93, opened 2019-09-05
Add setReturnFocus after deactivation option

jaakritso:
This option is useful when it is needed to set the returning element after deactivation programmatically.

In my case, another 3rd party library for modals sets the focus on the modal container. As I'm using the focus trap on modal, then the modal container focusing is called before focus trap sets the `state.nodeFocusedBeforeActivation`, which means after closing the modal focus trap won't return to previously focused element.

With this option I can manually set where it should return.

### DIFF
--- a/index.js
+++ b/index.js
@@ -75,7 +75,7 @@ function focusTrap(element, userOptions) {
 
     state.active = true;
     state.paused = false;
-    state.nodeFocusedBeforeActivation = getReturnFocusNode(doc.activeElement);
+    state.nodeFocusedBeforeActivation = doc.activeElement;
 
     var onActivate =
       activateOptions && activateOptions.onActivate
@@ -114,7 +114,7 @@ function focusTrap(element, userOptions) {
         : config.returnFocusOnDeactivate;
     if (returnFocus) {
       delay(function() {
-        tryFocus(state.nodeFocusedBeforeActivation);
+        tryFocus(getReturnFocusNode(state.nodeFocusedBeforeActivation));
       });
     }
 
@@ -221,18 +221,7 @@ function focusTrap(element, userOptions) {
 
   function getReturnFocusNode(previousActiveElement) {
     var node = getNodeForOption('setReturnFocus');
-
-    if (!node) {
-      node = previousActiveElement;
-    }
-
-    if (!node || !previousActiveElement) {
-      throw new Error(
-        'Your focus-trap needs to have at least one focusable element'
-      );
-    }
-
-    return node;
+    return node ? node : previousActiveElement;
   }
 
   // This needs to be done on mousedown and touchstart instead of click
--- a/index.js
+++ b/index.js
@@ -75,7 +75,7 @@ function focusTrap(element, userOptions) {
 
     state.active = true;
     state.paused = false;
-    state.nodeFocusedBeforeActivation = doc.activeElement;
+    state.nodeFocusedBeforeActivation = getReturnFocusNode(doc.activeElement);
 
     var onActivate =
       activateOptions && activateOptions.onActivate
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -219,6 +219,23 @@ function focusTrap(element, userOptions) {
     return node;
   }
 
+  function getReturnFocusNode(previousActiveElement) {
+    var node;
+    if (getNodeForOption('setReturnFocus') !== null) {
```
jtomaszewski: Can we extract `getNodeForOption('setReturnFocus')` to a variable in a line before it, so the `getNodeForOption` is called only once, not twice, in case it returns something?

index.js
```diff
@@ -219,6 +219,23 @@ function focusTrap(element, userOptions) {
...
+      node = getNodeForOption('setReturnFocus');
+    } else {
+      node = previousActiveElement;
+    }
+
+    if (!node) {
+      throw new Error(
+        "You can't have a focus-trap without at least one focusable element"
```
jtomaszewski: Badly worded if I'm correct, probably better would be to avoid double negation here like:

```suggestion
        "Your focus-trap needs to have at least one focusable element"
```

jtomaszewski:
Overall ok, I can see the use case. I'm not sure about the `setReturnFocus` naming but I don't have any better idea for it atm, so IMO it can stay as it is.

index.js
jaakritso: Refactored it.

jaakritso:
> Overall ok, I can see the use case. I'm not sure about the `setReturnFocus` naming but I don't have any better idea for it atm, so IMO it can stay as it is.
> 
> Thanks for the PR. Can you submit the corrections for suggestions? I think we could merge it afterwards, WDYT @davidtheclark ?

Thanks for the feedback. Refactored the code based on your suggestions. I agree that the `setReturnFocus` is not the clearest name, but I also couldn't come up with anything better so far. The problem with this naming is the conflict with another boolean option - `returnFocusOnDeactivate` and that's why I used the "set" prefix.

index.js
```diff
@@ -75,7 +75,7 @@ function focusTrap(element, userOptions) {
 
     state.active = true;
     state.paused = false;
-    state.nodeFocusedBeforeActivation = doc.activeElement;
+    state.nodeFocusedBeforeActivation = getReturnFocusNode(doc.activeElement);
```
jtomaszewski: Hmm, however shouldn't it be more obvious for `getReturnFocusNode` to be called in the moment when focus trap is being deactivated, instead of when focus trap is being activated? (See line 117)

index.js
```diff
@@ -212,7 +212,23 @@ function focusTrap(element, userOptions) {
...
+  function getReturnFocusNode(previousActiveElement) {
+    var node = getNodeForOption('setReturnFocus');
+
+    if (!node) {
+      node = previousActiveElement;
+    }
+
+    if (!node || !previousActiveElement) {
+      throw new Error(
+        'Your focus-trap needs to have at least one focusable element'
```
jtomaszewski: Sorry for my previous feedback. this error should tell something else, like "`document.activeElement` or `setReturnFocus` needs to return an element in order for the focus to be returned when focus-trap is deactivated".

But actually I'm not sure if we'd like to throw anything in here. I think in case in `node` is empty, we should just do nothing (then focus won't be returned). I mean, in a rare case, when `document.activeElement` is empty, and somebody programitically activates the focus trap, then if that trap is deactivated, then `document.activeElement` should still be empty, instead of having the app throw a fatal error, right?

index.js
jaakritso: Yeap. I was thinking as the `setReturnFocus` overwrites the returning element, then it does not matter. But, you're right, in case some event (if not now, maybe in future) needs `state.nodeFocusedBeforeActivation` before deactivation, then it makes sense to do it during deactivation.

I will change that.

index.js
jaakritso: Yeah, my bad. I used the `getInitialFocusNode()` function as a base to build `getReturnFocusNode`. I simplified the function with the last commit.

jaakritso:
@jtomaszewski any idea when this gets merged? Or should @davidtheclark check this out first? Thanks.

jtomaszewski:
I'm pretty sure it's good enough that we can merge it. But @davidtheclark is the only one that can release it (to npm).

davidtheclark:
@jtomaszewski I realize this has been sitting for a little while ... do you *want* it released? Wasn't sure from the last comment. If so, I can update the changelog and release within a day or two.

If you'd like to add 2FA to your npm account, I can add you as an npm collaborator, as well, for future releases.

jtomaszewski:
Yeah, should be fine.

Ill setup 2fa after the weekend ;)

davidtheclark:
Published in 5.1.0.

PR #94, opened 2019-09-25
Add all-inclusive ES6 module

TpmKranz:
This is my solution to https://github.com/material-components/material-components-web/issues/4507.

Since I plan on using mdc in a plain ES6 environment, I made this PR so that mdc may adopt the new version to enable just that.

In terms of output, only the esm file and its source map are new, the rest of the files should stay the same.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 **/*-bundle.js
 dist
 package*.json
+rollup.config.js
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,63 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.5.5",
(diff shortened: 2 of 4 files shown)


jtomaszewski:
Well they have been changed slightly:

image

The npm deps are now inlined in the dist file. Which is o'right in this case I guess?

(PR looks great.)

dy:
Any chance that will get merged? Having the same issue with importing material-components-web.

stefcameron:
@TpmKranz Thanks for this PR! I'm a new maintainer, trying to catch-up on things here. 😄  `tabbable` provides an ESM build with rollup similar to your solution here, so I'm going to see if I can update this branch after all the changes to master since, and take it in.

This would fix #30

stefcameron:
@TpmKranz Your fork was quite old and I had already made a lot of changes to master, so I followed what you proposed here and got it in as #139. It will go out in the next major: 6.0.0. Thanks again!

Issue #95, opened 2019-10-22
Option to not change focus on activation


dougalg:
I have a component which is dropdown list. So, there is a button to open/close the dropdown, and then a list of buttons inside it. I wrap the whole thing in a focus trap so that you can loop through all the option buttons and also the toggle button in the focus-trap.

However, I don't want to force focus anywhere when opening the trap, since the focus is already inside the focus trap. Can we add an option to allow this behaviour? I think there are 2 ways it can work

Option 1. Add a new option like `disableFocusOnActivate` which works as described
Option 2. When activating a focus trap, if focus is already inside it, do not move focus elsewhere

I'm happy to make a PR for this if we can agree on behaviour.

davidtheclark:
> I don't want to force focus anywhere when opening the trap, since the focus is already inside the focus trap

If focus is really already inside the trap, it should not be automatically moved. Is it possible you're creating a new element when the dropdown opens?

dougalg:
So, I looked into our code some more, and it turns out that I was mistaken. For our use case, we'd like to use focus-trap for mouse users to allow us to get "on outside click" support, but not force the focus to a specific element. But for keyboard users, we'd maintain the normal behaviour. For mouse users, it's possible to activate a focus trap while outside of the focus trap.

Would this likely be problematic for other reasons?

stefcameron:
Closing issues older than 2020. Please LMK if this is still a problem.

PR #96, opened 2019-10-22
Add focusOnActivate option

dougalg:
If set to false, activate() will not force focus on an element

Resolves https://github.com/davidtheclark/focus-trap/issues/95

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,6 +44,7 @@ function focusTrap(element, userOptions) {
 
   var config = xtend(
     {
+      focusOnActivate: true,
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -44,6 +44,7 @@ function focusTrap(element, userOptions) {
 
   var config = xtend(
     {
+      focusOnActivate: false,
```
dougalg: alternately, this could be in the `activateOptions`

stefcameron:
Closing as #95 has been closed.

PR #98, opened 2019-10-30
Add `document` option

### LINKED ISSUE #97: Focus-trap inside iframe
**Context:**
 - Render DOM from the top window runtime inside an iframe
 - Then, activate the focus-trap on an rendered element

**Issue encountered:**
The referred document is the top window document and not the iframe inner document.

MD4:
Default: `window.document`. Document where the focus trap will be active.

Resolves #97


stefcameron:
Closing PRs older than 2020. Happy to re-open if tests can be added, and more context about the problem given.

PR #101, opened 2019-11-12
feat: if not active, don't try focusing

cgood92:
Fixing issue where deactivation of focus trap could happen before attempting to focus something again on a delay.

This will fix a problem we are having.  Here is a replication of that problem: https://codesandbox.io/s/upbeat-mendel-kbdzw .

### DIFF
--- a/index.js
+++ b/index.js
@@ -301,6 +301,7 @@ function focusTrap(element, userOptions) {
   }
 
   function tryFocus(node) {
+    if (!state.active) return;
     if (node === doc.activeElement) return;
     if (!node || !node.focus) {
       tryFocus(getInitialFocusNode());


stefcameron:
@cgood92 Thanks for this PR. Sorry it's taken so (very) long to have someone look at it. The repo fell out of maintenance for the better part of the last 12 months, and I've just taken over a couple of months ago, and trying to catch-up.

Thanks for the sandbox. I forked it at https://codesandbox.io/s/hopeful-field-2i9fr in order to update the `focus-trap-react` and `focus-trap` dependencies to the latest versions to make sure this was still an issue.

As best I can tell, after staring at this for a long while, is that it's actually related to what https://github.com/focus-trap/focus-trap-react/pull/54 is trying to fix, which is essentially a double-deactivation loop with the focus trap that happens when `clickOutsideDeactivates=true`, because with that option enabled, `focus-trap` immediately deactivates the trap when you click on the green square (in the sandbox), and then the `<FocusTrap/>` component unmounts and attempts to deactivate it again, and things to bad, similar to the issue described in that PR.

Unfortunately, the fix to `returnFocusOnDeactivate` in `focus-trap@6.1.1` didn't help.

index.js
```diff
@@ -301,6 +301,7 @@ function focusTrap(element, userOptions) {
   }
 
   function tryFocus(node) {
+    if (!state.active) return;
```
stefcameron, line 304: The fix you're proposing here seems right, but I don't like that we're checking for the activate state in `tryFocus()`. I think the fix belongs more in the timeout related to the delayed focus on activation, [at line 145](https://github.com/focus-trap/focus-trap/blob/master/index.js#L145), just before calling `tryFocus()`. 

`tryFocus()` itself doesn't need to be concerned about state.

And with newer options in `focus-trap`, I thought the new `delayInitialFocus` option set to `false` might help since your description makes it sound like that's basically the problem (because you say the trap is deactivated by the time the initial focus is set), but that didn't help either. Maybe I misinterpreted something there.

stefcameron (changes requested):

index.js
stefcameron: @cgood92 Since you're around... any thoughts on this one?

stefcameron: Cool, thanks!

PR #120, opened 2020-08-10
feat: Add boolean value support for allowOutsideClick option

SeanMcP:
## Description

This adds support for boolean value to be passed to the `allowOutsideClick` option. This addresses focus-trap/focus-trap-react#86 and is the follow-up to focus-trap/focus-trap-react#87.

### Documentation

When describing the type of `allowOutsideClick`, I went with "If set and is or returns `true`". There may be a better wording for this.

### Demo

I wanted to add an example without cluttering the page with another example that is visually identical. I opted for a `select` element to switch between passing a function or a boolean. This is contained in ceec9da.

### DIFF
--- a/demo/js/allowoutsideclick.js
+++ b/demo/js/allowoutsideclick.js
@@ -3,20 +3,21 @@ var createFocusTrap = require('../../');
 var container = document.getElementById('allowoutsideclick');
 var trigger = document.getElementById('activate-allowoutsideclick');
 var active = false;
+var allowOutsideClick = true;
 
-var focusTrap = createFocusTrap('#allowoutsideclick', {
-  allowOutsideClick: function(event) {
-    if (event.target === trigger) {
-      return true;
+function initialize() {
+  return createFocusTrap('#allowoutsideclick', {
+    allowOutsideClick: allowOutsideClick,
+    onActivate: function() {
+      container.className = 'trap is-active';
+    },
+    onDeactivate: function() {
+      container.className = 'trap';
     }
-  },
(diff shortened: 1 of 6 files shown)


SeanMcP:
@stefcameron Done in f0f6133

demo/js/allowoutsideclick.js
```diff
@@ -3,20 +3,21 @@ var createFocusTrap = require('../../');
...
 var active = false;
+var allowOutsideClick = true;
 
-var focusTrap = createFocusTrap('#allowoutsideclick', {
-  allowOutsideClick: function(event) {
-    if (event.target === trigger) {
-      return true;
+function initialize() {
+  return createFocusTrap('#allowoutsideclick', {
+    allowOutsideClick: allowOutsideClick,
```
stefcameron, line 10: This can be shortened to just `allowOutsideClick`. :)

SeanMcP: Good catch! I had that initially, but there was a parsing error when I ran `npm run lint`:

```
/Users/seanmcp/dev/seanmcp/focus-trap/demo/js/allowoutsideclick.js
  10:22  error  Parsing error: Unexpected token ,

✖ 1 problem (1 error, 0 warnings)
```

I couldn't figure out what was wrong, and it didn't seem worth the stress for a demo.

stefcameron: No worries. I haven't updated all the deps and tooling on this repo like I have on the react one so there's probably some babel-eslint thing missing.

PR #123, opened 2020-08-12
ts: fix TypeScript syntax for allowOutsideClick option

michael-ar:
[Recent PR](https://github.com/focus-trap/focus-trap/pull/120) contains invalid syntax for new option, this fixes.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,6 +5,8 @@ declare module 'focus-trap' {
    * returns a DOM node.
    */
   export type FocusTarget = HTMLElement | string | { (): HTMLElement };
+  
+  type MouseEventToBoolean = (event: MouseEvent) => boolean
 
   export interface Options {
     /**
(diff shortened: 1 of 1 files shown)


stefcameron:
@michael-ar Thanks for spotting that! TS is not my forte, though I would never have thought `boolean | () => boolean` was invalid syntax, but indeed it is.

michael-ar:
@stefcameron all good — I would have intuitively thought so too!

do you have any idea when the next npm release might happen? am currently installing directly from GitHub as I need one of the unreleased options

stefcameron:
> @stefcameron all good — I would have intuitively thought so too!

😅  What's the best way to verify the typings? Run the file through the TS compiler somehow, or through [typescript-eslint](https://github.com/typescript-eslint/typescript-eslint)? I'd like to add some simple tooling to validate the syntax as part of the quality process.

> do you have any idea when the next npm release might happen? am currently installing directly from GitHub as I need one of the unreleased options

I'm currently working on #124 and then will publish an update, most likely 6.0.0 just to be safe because of the path change to `main` in package.json, though I don't anticipate any breaking changes at all. I hope to publish sometime within the next week.

stefcameron:
> What's the best way to verify the typings?

I figured out it's just to add `typescript` and then `tsc index.d.ts`, so that's what I've got going on in #125

michael-ar:
> > What's the best way to verify the typings?
> 
> I figured out it's just to add `typescript` and then `tsc index.d.ts`, so that's what I've got going on in #125

Oh nice one, definitely nice to have as part of the automated flow. Thanks again!

SeanMcP:
> @michael-ar Thanks for spotting that! TS is not my forte, though I would never have thought `boolean | () => boolean` was invalid syntax, but indeed it is.

@stefcameron That makes three of us! I'll file that away for next time.

PR #125, opened 2020-08-15
Update all dependencies, move to yarn/Webpack and GitHub CI

### LINKED ISSUE #124: Refresh all dependencies and move to GH workflow for CI
@maraisr I dove into this today and thought I better let you know so we don't double efforts. I'm not doing anything WRT Changesets; I'll leave that to you, but might be best to wait until I fix this issue.

I'm updating _everything_ the same as we did for `focus-trap-react` with ESLint, Prettier, Babel, Yarn, etc.

I'm also fixing the fact that `main: "index.js"` in package.json should be `main: "dist/focus-trap.js"`.

With all the deps updated to latest, it should result in closing a bunch of the security-related Dependabot issues.

Once that's done, I hope we can release 5.1.1 (or maybe 6.0.0 just to be safe with all these changes, especially in light of the path change above) with the pending unreleased changes, especially for https://github.com/focus-trap/focus-trap-react/pull/87

stefcameron:
Fixes #124 to put this repo inline with focus-trap-react:

- update all deps to latest versions
- move to yarn with yarn.lock
- use Webpack to build instead of Browserify (see below
  for more details)
- update Prettier and run on all the source (same rules as
  focus-trap-react)
- add check for types so that problems like #123 don't
  happen again
- (*) fixes bug in package.json 'main' that wasn't pointing to
  /dist/focus-trap.js (build output)

Because of (*) this should be published as a new __major__ release,
just to be safe, even though there shouldn't be any breaking
changes as a result (famous last words...).

This should also close-out a bunch of the security-related
notices on the repo from Dependabot because of old package
versions with vulnerabilities.

Added Webpack builds and removed some deps:

- Webpack Dev and Prod builds, mimicing old Browserify builds.
- Removed Browserify: Webpack replaces.
- Removed uglify: Webpack takes care of this now.
- Removed xtend: simple ES6 object merging works fine.

Fixed 2 small bugs:

- Fixed bug on line 316 of index.js where an NPE would happen
  if the trap was created without `userOptions`; should be
  `config.preventScroll` instead since we create `config` early
  on within the module and use that everywhere instead of
  `userOptions`.
- Fixed bug in allow-click-outside demo by disabling ESC: Otherwise,
  the 'activate trap' button state would get out of sync with
  the state of the trap (would remain 'deactivate trap' after
  pressing ESC).

NOTE: Publishing is still done manually from a local repo.
If we move this repo to Changesets later, we can start
doing releases via GitHub Actions like `focus-trap-react`.


stefcameron:
@maraisr 
> WoW nice one mate, yeah will do the whole changesets thing. And maybe simplify our build too

Thanks! Sounds good. What I did here, I will now do over on `tabbable`. Then we'll have the whole "family" up to date.

PR #139, opened 2020-09-04
Move to rollup and provide an ESM build

### LINKED ISSUE #30: Not compatible with ES2015 / ES6 (import)

stefcameron:
Fixes #30

BIG THANKS to @TpmKranz for #94 which I used as a template for this
since the #94 branch was too old WRT current master.


stefcameron:
FYI, this should fix https://github.com/material-components/material-components-web/issues/4507 when we publish focus-trap 6.0.0 shortly. 😄

Issue #143, opened 2020-09-05
Leverage Cypress to add coverage for all/future demos


stefcameron:
Unit tests with JSDom aren't effective because JSDom doesn't fully support focus events. Cypress is much better suited.

The goal is to add Cypress E2E tests to cover each demo's expected functionality, and to add new demos as test cases to cover future API enhancements.

Run E2E testing at each `push` and `pull-request` on major browsers:

- [x] Chrome
cypress supported

- [x] Firefox
cypress supported
Having problem running on git action with error msg `Browser: 'firefox' was not found on your system or is not supported by Cypress`, see [cypress image open issue](https://github.com/cypress-io/cypress-docker-images/issues/363)

- [ ] Edge(chromium-based)
cypress supported

- [ ] Edge(EdgeHTML/pre-chromium/IE11?)
Not supported by cypress at this moment. See [cypress IE 11 support github issue](https://github.com/cypress-io/cypress/issues/6423)

- [ ] Safari
Not supported by cypress at this moment. See [cypress macOS Safari support github issue](https://github.com/cypress-io/cypress/issues/6422)

FYI, this is also what we do over on `focus-trap-react`: https://github.com/focus-trap/focus-trap-react/issues/104

liunate:
I can help some cypress tests over the weekend, how about let me take these tests?

```javascript
  describe('demo: iene', () => {
    // TODO
  });

  describe('demo: ifc', () => {
    // TODO
  });

  describe('demo: ht', () => {
    // TODO
  });
```

stefcameron:
@liunate Certainly! Please do. Much appreciated!

stefcameron:
#166 adds official FireFox support to Cypress! 🎉

stefcameron:
We seem to be doing pretty good with testing on Chrome and FF, and those seem to be the main ones that Cypress supports. Closing.

PR #144, opened 2020-09-05
Add Cypress e2e tests against the demos

stefcameron:
Since 'the demo is also the test' in this repo, let's use the
advent of Cypress (which is able to effectively test things
related to focus, unlike unit tests with JSDom, which doesn't
fully support focus events) to automate running all the demos
and exercise the focus-trap API.

NOTE: `demo/index.html` is UNCHANGED in functionality. I just
added a wrapper div with ID to each demo to make it easier
to find each demo when running Cypress tests.

The framework is in place, and the first demo is tested. Other
demos are TODO. I'm hoping to get community help to complete them.
See #143

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,3 +50,23 @@ jobs:
 
       - name: Build # Tests to see if a build can succeed
         run: yarn build
+  e2e:
+    runs-on: ubuntu-latest
+    container: cypress/included:5.0.0
+    strategy:
+      matrix:
+        # cypress executable says 'Browser: 'firefox' was not found on your system
+        #  or is not supported by Cypress.' seems to be defect
+        #  @see https://github.com/cypress-io/cypress-docker-images/issues/363
+        browser: [chrome]
+    env:
+      CYPRESS_BROWSER: ${{ matrix.browser }}
+    steps:
+      - name: Checkout 🛎
+        uses: actions/checkout@master
+
+      - name: Install dependencies 📦
+        uses: bahmutov/npm-install@v1
+
+      - name: Run tests 🧪
+        run: yarn test-cypress-ci
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ---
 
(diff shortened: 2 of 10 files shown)


stefcameron:
No changeset needed. This is just tooling.

liunate:
It seems `focus-trap-react` and its base package(this `focus-trap` repo) sharing mostly identical test cases? 

If that's the case, do we consider making these two repos a monorepo that publishes their own packages while sharing some developer scripts and same versioning? 

I'd imagine, for test cases particularly, `focus-trap` will have most crucial test cases while `focus-trap-react` has some but not many react specific test cases.

stefcameron:
> It seems `focus-trap-react` and its base package(this `focus-trap` repo) sharing mostly identical test cases?

I wouldn't say identical. They shouldn't be. Test cases in focus-trap-react should focus mostly on the React-specific API of active/paused, activate/deactivate on mount/unmount, etc., as well as just ensuring that focus is properly trapped, as expected (but on a basic level, not deep into all possible combinations of focus-trap settings -- unless some of those settings have nuances in the React behavior).

focus-trap test cases should focus on all the different trap options and how they interact with one another.

> If that's the case, do we consider making these two repos a monorepo that publishes their own packages while sharing some developer scripts and same versioning?

Moving to a monorepo is a long-term goal. Since I took over as a maintainer, I've been working to get all the tech updated, and I've been aligning all the tools like rollup for builds, etc., across all 3 repos (don't forget about tabbable).

Ultimately, it would be nice to converge on this repo, `focus-trap/focus-trap` as the monorepo for all 3 repos currently in the org, since they're all part of the same family. I'd even consider moving the packages to `@focus-trap/tabbable | focus-trap | focus-trap-react` on NPM, but we'd have to first have a plan to post package name deprecation notices on `npm install` for a while before we do that.

> I'd imagine, for test cases particularly, `focus-trap` will have most crucial test cases while `focus-trap-react` has some but not many react specific test cases.

As I mentioned above, focus-trap-react should focus on React-specific behavior more than trying to test all possible focus-trap options.

It's OK if there's a bit of overlap, though, and I would expect there to be some, and also, even if we had a monorepo today, each package would still have its own specific tests anyway. I don't see a monorepo as being a "solution" to not duplicating tests (to a certain degree).

In my mind, I already treat all the repos as a monorepo. They're just not in a single repo because that's not how they started off. 🙂

PR #149, opened 2020-09-15
Some cypress tests

liunate:
Refer back to #143 Add some cypress tests against current demo pages. 

###  Features and Bug Fixes

- [ ] ~~Typings added/updated.~~
- [ ] ~~README updated (API changes, instructions, etc.).~~
- [ ] ~~Changes to dependencies explained.~~
- [ ] ~~Changeset added (run `yarn changeset` locally to add one, follow prompts).~~

### Tooling

- [ ] ~~Issue being fixed is referenced.~~
- [ ] ~~Changes to dependencies explained.~~
- [ ] ~~README instructions updated.~~
- [x] A Changeset is __not__ added.

👉 If any of these don't apply, please ~cross them out~.

### DIFF
--- a/cypress/integration/focus-trap-demo.spec.js
+++ b/cypress/integration/focus-trap-demo.spec.js
@@ -108,11 +108,13 @@ describe('focus-trap', () => {
   describe('demo: iene', () => {
     beforeEach(() => {
       cy.get('#demo-iene').as('testRoot');
-    })
+    });
 
     it('On trap activation, focus on manually specified input element', () => {
       // activate trap
-      cy.get('@testRoot').findByRole('button', { name: 'activate trap' }).click();
+      cy.get('@testRoot')
+        .findByRole('button', { name: 'activate trap' })
+        .click();
 
       // instead of next tab-order element being focused, element specified should be focused
       cy.get('@testRoot')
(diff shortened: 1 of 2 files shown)


cypress/integration/focus-trap-demo.spec.js
```diff
@@ -1,6 +1,18 @@
 describe('focus-trap', () => {
   beforeEach(() => cy.visit('index.html'));
 
+  function verifyCrucialFocusTrapOnClicking(focusedElAlias) {
```
liunate, line 4: Some very basic clicking test to ensure trap is still effective

demo/index.html
```diff
@@ -63,7 +63,7 @@ <h2 id="iene-heading">initial element, no escape</h2>
         Here is a focus trap <a href="#">with</a> <a href="#">some</a> <a href="#">focusable</a> parts.
       </p>
       <p>
-        <label htmlFor="focused-input" class="inline-label">
+        <label for="focused-input" class="inline-label">
```
liunate, line 66: `htmlFor` is not a valid attribute for `<label>` but instead `for` should be used so the cypress accessibility 
 selector `findByRole` works

demo/index.html
```diff
@@ -79,8 +79,8 @@ <h2 id="iene-heading">initial element, no escape</h2>
   <div id="demo-ifc">
     <h2 id="ifc-heading">initially focused container</h2>
     <p>
-      Initial focus is on the containing element, which has <code>tabindex="-1"</code> and no focus outline;
-      so when you tab through the trap focus does not return to the container.
+      When this focus trap activates, initial focus is on the containing element(which has <code>tabindex="-1"</code> so is not tabbable),
+      so when you tab through the trap, focus does not return to the containing element.
```
liunate, line 83: I feel previous version can be easier to understand by some rewording.

demo/index.html
stefcameron: For sure. `htmlFor` is the React equivalent because `for` is seen as a JS statement in JSX. Force of habit for whoever put that in there. 😉  And good catch by you!

demo/index.html
stefcameron: What would you think about wording it this way?
```
When this focus trap activates, initial focus is on the containing element (which has <code>tabindex="-1"</code> and is therefore not tabbable).
When you tab through the trap, focus does not return to the containing element.
```
Just a _slight_ tweak. 🙂

stefcameron:
@liunate Thanks for adding more tests! Much appreciated.

demo/index.html
liunate: Looks better than my version

stefcameron (approved):

PR #154, opened 2020-09-18
Add delayInitialFocus option

sadick254:
A change was introduce to add delay before autofocus in this PR https://github.com/focus-trap/focus-trap/pull/51 . The example given was https://codepen.io/anon/pen/RyvdEZ?editors=1111. Looking at the example, the bug described in the issue is not a `focus-trap` bug. 

The desired effect @pgn-vole was hopping to achieve could be done by calling `e.preventDefault()` on the event handler on the example instead of using `setTimeout`.

This change has not allowed us to upgrade to the latest version `6.0.1` because the behaviour of `createFocusTrap` changed without any additional documentation.

We are using this awesome library in the [atom text editor](https://github.com/atom/atom/pull/21256) 

###  Features and Bug Fixes

~- [] Test coverage added/updated.~
~- [ ] Typings added/updated.~
~- [ ] README updated (API changes, instructions, etc.).~
~- [ ] Changes to dependencies explained.~
~- [ ] Changeset added (run `yarn changeset` locally to add one, follow prompts).~

👉 If any of these don't apply, please ~cross them out~.

### Tooling

~- [ ] Issue being fixed is referenced.~
~- [ ] Changes to dependencies explained.~
~- [ ] README instructions updated.~
~- [ ] A Changeset is __not__ added.~

👉 If any of these don't apply, please ~cross them out~.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,8 @@ Returns a new focus trap on `element`.
 - **allowOutsideClick** {boolean|(e: MouseEvent) => boolean}: If set and is or returns `true`, a click outside the focus trap will not be prevented, even when `clickOutsideDeactivates` is `false`. When `clickOutsideDeactivates` is `true`, this option is **ignored** (i.e. if it's a function, it will not be called). Use this option to control if (and even which) clicks are allowed outside the trap in conjunction with `clickOutsideDeactivates: false`.
 - **returnFocusOnDeactivate** {boolean}: Default: `true`. If `false`, when the trap is deactivated, focus will *not* return to the element that had focus before activation.
 - **setReturnFocus** {element|string|function}: By default, focus trap on deactivation will return to the element that was focused before activation. With this option you can specify another element to programmatically receive focus after deactivation. Can be a DOM node, or a selector string (which will be passed to `document.querySelector()` to find the DOM node), or a function that returns a DOM node.
-- **preventScroll** {boolean}: By default, focus() will scroll to the element if not in viewport. It can produce unintended effects like scrolling back to the top of a modal. If set to `true`, no scroll will happen. 
+- **preventScroll** {boolean}: By default, focus() will scroll to the element if not in viewport. It can produce unintended effects like scrolling back to the top of a modal. If set to `true`, no scroll will happen.
+- **delayInitialFocus** {boolean}: Default: `true`. Delays the autofocus when the focus trap is activated. This prevents elements within the focusable element from capturing the event that triggered the focus trap activation.
 
 ### focusTrap.activate([activateOptions])
 
--- a/cypress/integration/focus-trap-demo.spec.js
+++ b/cypress/integration/focus-trap-demo.spec.js
@@ -230,7 +230,35 @@ describe('focus-trap', () => {
   });
 
   describe('demo: delay', () => {
-    // TODO
+    it('activates the focus trap when delayInitialFocus is set to true', () => {
+      cy.get('#delay').should(($div) => {
+        expect($div[0].style.opacity).to.equal('0.2');
+      });
+      cy.get('#activate-delay').type('{enter}');
+      cy.get('#delay').should('have.class', 'trap is-active');
+      cy.get('#delay').should(($div) => {
+        expect($div[0].style.opacity).to.equal('1');
+      });
+      cy.get('#close-button-delay').as('focusedEl').should('have.focus');
+
+      // crucial focus-trap feature: mouse click is trapped
+      verifyCrucialFocusTrapOnClicking('@focusedEl');
+    });
+
+    it('activates the focus trap when delayInitialFocus is set to false', () => {
+      cy.get('#no-delay').should(($div) => {
+        expect($div[0].style.opacity).to.equal('0.2');
+      });
+      cy.get('#activate-no-delay').type('{enter}');
+      cy.get('#no-delay').should('have.class', 'trap is-active');
+      cy.get('#no-delay').should(($div) => {
+        expect($div[0].style.opacity).to.equal('1');
+      });
+      cy.get('#close-button-no-delay').as('focusedEl').should('have.focus');
+
+      // crucial focus-trap feature: mouse click is trapped
+      verifyCrucialFocusTrapOnClicking('@focusedEl');
+    });
   });
 
   describe('demo: radio', () => {
--- a/demo/js/no-delay.js
+++ b/demo/js/no-delay.js
@@ -1,0 +1,33 @@
+var { createFocusTrap } = require('../../dist/focus-trap');
+
+var container = document.getElementById('no-delay');
+
+var focusTrap = createFocusTrap(container, {
+  delayInitialFocus: false,
+  onActivate() {
+    container.style.opacity = '1';
+    container.classList.add('is-active');
+  },
+  onDeactivate() {
+    container.style.opacity = '0.2';
+    container.classList.remove('is-active');
+  },
+});
+
+document
+  .getElementById('activate-no-delay')
+  .addEventListener('keydown', showContainer);
+document
+  .getElementById('close-button-no-delay')
+  .addEventListener('click', hideContainer);
+
+function showContainer(e) {
+  if (e.keyCode === 13) {
+    e.preventDefault();
+    focusTrap.activate();
+  }
+}
+
+function hideContainer() {
+  focusTrap.deactivate();
+}
--- a/index.d.ts
+++ b/index.d.ts
@@ -72,6 +72,12 @@ declare module 'focus-trap' {
      * If set to `true`, no scroll will happen.
      */
     preventScroll?: boolean;
+    /**
+     * Default: `true`. Delays the autofocus when the focus trap is activated.
+     * This prevents elements within the focusable element from capturing
+     * the event that triggered the focus trap activation.
+     */
+    delayInitialFocus?: boolean;
   }
 
   type ActivateOptions = Pick<Options, 'onActivate'>;
--- a/index.js
+++ b/index.js
@@ -44,6 +44,7 @@ function createFocusTrap(element, userOptions) {
   var config = {
     returnFocusOnDeactivate: true,
     escapeDeactivates: true,
+    delayInitialFocus: true,
     ...userOptions,
   };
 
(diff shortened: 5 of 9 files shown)


stefcameron:
@sadick254 Thanks for posting this PR. Please bare with me as I continue to get up to speed on `focus-trap`'s history now that I've taken over as maintainer since last July or so.

> A change was introduce to add delay before autofocus in this PR #51 . The example given was https://codepen.io/anon/pen/RyvdEZ?editors=1111. Looking at the example, the bug described in the issue is not a `focus-trap` bug.
>
> The desired effect @pgn-vole was hopping to achieve could be done by calling `e.preventDefault()` on the event handler on the example instead of using `setTimeout`.

You are right that the delay was, in fact, not necessary. I can reproduce the bug with the delay demo that #51 added if I remove the timeout, and I can fix the demo by adding `e.preventDefault()` in [delay.js](https://github.com/focus-trap/focus-trap/blob/master/demo/js/delay.js#L25) at line 25.
 
> This change has not allowed us to upgrade to the latest version `6.0.1` because the behaviour of `createFocusTrap` changed without any additional documentation.

What I see in the repository is that #51 was merged on June 10, 2018 (over 2 years ago) and the change promptly published in 2.4.6, with a comment in the CHANGELOG: https://github.com/focus-trap/focus-trap/blob/master/CHANGELOG.md#246

That's basically the level of documentation you would've expected from this repository at the time. Not sure what to say there, I wasn't around back then.

Subsequently, there were 3 other major releases, up to 5.1.0, which is the final release the previous maintainer (and author) made: https://github.com/focus-trap/focus-trap/blob/master/CHANGELOG.md#510

I took over and released 6.0.0 and 6.0.1, but the delay had long been in the library. It's been in there for over 2 years now.

At first, I thought I might have inadvertently been the one to publish it without knowing the implications when I took over and published 6.0.0, and then we could just call this a bug fix, but there's over 2 years of precedent with the behavior.

And since I would look at this as being a significant change in behavior, we'd have to call this a new major release. But I'm concerned it could also break expectations that other consumers have come to establish with `focus-trap` behavior over the past 2 years, even though it's technically not necessary. Maybe there are other cases where this actually helps.

> We are using this awesome library in the [atom text editor](https://github.com/atom/atom/pull/21256)

I'm definitely up for trying to make it right, but I think we need to be cautious and not just revert the change like you're suggesting. I think at this point, because of the history, we need to put this behind a flag by adding another option like `delayInitialFocus: true` (i.e. true -- the current behavior -- by default).

Then we could publish this as a minor release, and you could upgrade to 6.1.0 and set `delayInitialFocus: false` in your trap options.

What do you think?

sadick254:
@stefcameron That sounds like a plan. I have made the change. 👍

index.js
```diff
@@ -319,6 +319,10 @@ function createFocusTrap(element, userOptions) {
       node.select();
     }
   }
+
+  function delay(fn) {
```
stefcameron, line 323: `delay()` is also called at line 113 as part of the deactivation process. This change would affect that also.

Plus, here, we're defaulting to `false`, which means no delay, while we need to default to `true`, which means to delay, thus preserving the existing behavior.

stefcameron (changes requested):
@sadick254 I'm glad you agree with my suggestion to add the `delayInitialFocus` option. Thanks for starting the change!

Aside from the issue I pointed out below, here's what this will take:

- [ ] Add documentation for the new `delayInitialFocus` option to the README, where all the other existing options are documented.
- [x] Make sure the option defaults to `true` (not `false`) and only affects __activation__ since we're talking about initial focus, not optional returned focus on deactivate. We should do something similar to [escapeDeactivates](https://github.com/focus-trap/focus-trap/blob/master/index.js#L46) with that defaults to `true` to setup the config...
- [ ] Add the new option to the `index.d.ts` typings.
- [ ] Add a new demo to test the option: My suggestion, to make this easy, is to duplicate the `delay.js` demo, but just set it up with the `preventDefault()` and the option set to `false`. The demo explanation should state what the difference is with the `delay` demo.
- [ ] Add test coverage for the new demo: This is done using Cypress since that's the only reliable way to test focus. See the existing tests in [focus-trap-demo.spec.js](https://github.com/focus-trap/focus-trap/blob/master/cypress/integration/focus-trap-demo.spec.js) and just add a `describe('no-delay')` (or whatever you call it) test for it that sets focus on the outer button, presses the ENTER key, and verifies that the inner button then has focus and the focus is trapped. You should be able to use the new [verifyCrucialFocusTrapOnClicking()](https://github.com/focus-trap/focus-trap/blob/master/cypress/integration/focus-trap-demo.spec.js#L4) helper we just merged from #149 to do that.
- [x] Add a changeset for a __minor__ update. Run `yarn changeset` and follow the prompts.

index.js
stefcameron: I would recommend we leave `delay()` where it's currently defined, without modifying it, and just an an IF statement around line 140 where the ELSE just calls `tryFocus(getInitialFocusNode());` immediately.

sadick254: @stefcameron Done.

cypress/integration/focus-trap-demo.spec.js
```diff
@@ -122,7 +122,29 @@ describe('focus-trap', () => {
...
+      cy.get('#no-delay').should(($div) => {
+        expect($div[0].style.opacity).to.equal('0.2');
+      });
+      cy.get('#activate-no-delay').type('{enter}');
+      cy.get('#no-delay').should('have.class', 'trap is-active');
+      cy.get('#no-delay').should(($div) => {
+        expect($div[0].style.opacity).to.equal('1');
+      });
+      cy.get('#close-button-no-delay').should('have.focus');
+    });
```
sadick254, line 147: I know this could be simplified. But not sure if that is something you would like.

README.md
```diff
@@ -78,7 +78,8 @@ Returns a new focus trap on `element`.
...
 - **setReturnFocus** {element|string|function}: By default, focus trap on deactivation will return to the element that was focused before activation. With this option you can specify another element to programmatically receive focus after deactivation. Can be a DOM node, or a selector string (which will be passed to `document.querySelector()` to find the DOM node), or a function that returns a DOM node.
-- **preventScroll** {boolean}: By default, focus() will scroll to the element if not in viewport. It can produce unintended effects like scrolling back to the top of a modal. If set to `true`, no scroll will happen. 
+- **preventScroll** {boolean}: By default, focus() will scroll to the element if not in viewport. It can produce unintended effects like scrolling back to the top of a modal. If set to `true`, no scroll will happen.
+- **delayInitialFocus** {boolean}: Delays the autofocus when the focus trap is activated. This prevents elements within the focusable element from capturing the event that triggered the focus trap activation.
```
stefcameron, line 82: Please update this to also mention it defaults to `true`.

demo/js/no-delay.js
```diff
@@ -0,0 +1,32 @@
+var { createFocusTrap } = require('../../dist/focus-trap');
+
+var container = document.getElementById('no-delay');
+
+var focusTrap = createFocusTrap(container, {
```
stefcameron, line 5: This "no-delay" trap should be setting the `delayInitialFocus: false` option in its configuration.

index.d.ts
```diff
@@ -72,6 +72,12 @@ declare module 'focus-trap' {
      * If set to `true`, no scroll will happen.
      */
     preventScroll?: boolean;
+    /**
+     * Delays the autofocus when the focus trap is activated.
+     * This prevents elements within the focusable element from capturing
+     * the event that triggered the focus trap activation.
```
stefcameron, line 78: Please mention default is `true` here also.

cypress/integration/focus-trap-demo.spec.js
stefcameron: I'm OK with what you've done there, but I would request that the tests also verify that the trap has been successfully activated by using the [verifyCrucialFocusTrapOnClicking()](https://github.com/focus-trap/focus-trap/blob/master/cypress/integration/focus-trap-demo.spec.js#L4) helper function to verify focus is now trapped as expected.

index.js
```diff
@@ -137,11 +138,14 @@ function createFocusTrap(element, userOptions) {
...
     // Delay ensures that the focused element doesn't capture the event
     // that caused the focus trap activation.
-    activeFocusDelay = delay(function () {
-      tryFocus(getInitialFocusNode());
-    });
+    activeFocusDelay = config.delayInitialFocus
+      ? delay(function () {
+          tryFocus(initialFocusNode);
+        })
+      : tryFocus(initialFocusNode);
```
stefcameron, line 148: While I definitely appreciate the optimization of calling `getInitialFocusNode()` once for either case of `delayInitialFocus`, and probably something I would have done myself, seeing it here is making me realize that the previous behavior was actually allowing the initial focus node to not exist in the current event loop pass that activates the trap because the call to `getInitialFocusNode()` was also delayed to the pass. That could potentially allow for the initial focus node to be created in the same pass that activates the trap, after calling `activate()` on the trap.

__TL/DR__: So just to be extra cautious, I'd prefer _not_ to optimize here, and call `getInitialFocusNode()` in both places rather than using the variable approach.

stefcameron (changes requested):
Thank you for diligently addressing all my requests for this new feature! What a huge help!

I had a look and found a couple of things I think should be updated. They should be easy to address.

Also, please delete the `cypress/videos/focus-trap-demo.spec.js.mp4` file that got committed (Cypress creates these recording files now). Files in `cypress/video` should be ignored. I think you might need to rebase master since `.gitignore` was updated very recently to ignore that directory.

I'll publish an update as soon as we have this merged to unblock you on the Atom side.

sadick254:
@stefcameron I have made all the changes requested. I have also cleaned up the commits. Take a final look at it and let me know if there is anything I have missed.

stefcameron (approved):
Fantastic! Thank you, this looks great! 💯

stefcameron:
I'm merging this! I'll circle back once I've done the release.

stefcameron:
NPM appears to be on the fritz, returning 500s. Not sure how long it'll take to publish, but publish is failing right now because of NPM.

stefcameron:
So now NPM says it's back up and running, but I can't publish because it keeps requiring an OTP but OTP is not required for the account. 😑  I published `tabbable` yesterday with the same account, worked fine, no OTP request...

stefcameron:
I finally figured out (with NPM's help) what the issue was and 6.1.0 is now published! 🎉 

Thanks again for your help. I hope this makes it now possible for you to bump Atom up to `focus-trap@6.1.0`.

Issue #172, opened 2020-10-05
IE11 not working with latest release


bparish628:
The issue here is that a new feature was added in v6 that does not include ie11 supported code. The culprit is the because the userOptions is being spread onto an object. 
https://github.com/focus-trap/focus-trap/blob/master/index.js#L48

Introduced in https://github.com/focus-trap/focus-trap/pull/125

Does this code get transpiled or should the src be compliant? I'd be happy to make a fix to it, just want to verify what the intention is!

Error is `Expected identifier, string or number`
@ this line
![image](https://user-images.githubusercontent.com/8492971/95091605-44211780-06ec-11eb-9d6d-263eda945ade.png)

Steps to Recreate

* Update to v6
* Run in ie11 and see js errors

stefcameron:
@bparish628 👋  Thanks for reaching out! We definitely [transpile](https://github.com/focus-trap/focus-trap/blob/master/babel.config.js) the code (and those settings, without any browser targets, _should_ be putting everything back down the ES5) -- at least, it's _supposed_ to be getting transpiled!

However, looking at `./dist/focus-trap.js` in the published package, I can see that it's in fact not the case, which isn't the intent.

I'll try to fix this ASAP!

stefcameron:
Problem seems to be that `@rollup/plugin-babel` is ignoring `babel.config.js` when the build script runs. Compiling `babel index.js` produces properly transpiled code that would be ES5-compatible.

stefcameron:
Unfortunately, I've been trying to get rollup (not a fan!!) to use Babel for the past hour and a half, trying everything I can think of, nothing is working. `@rollup/plugin-babel` does read the config (because it throws an error and fails the build if I include an invalid config option Babel doesn't know about), but does absolutely nothing with it. The code never gets transpiled. If I run `babel index.js`, as I mentioned before, I get the desired output. So the issue is with `@rollup/plugin-babel`... Any ideas?

stefcameron:
Of course, _no one on the Internet_ seems to have this issue. The plugin just works...

bparish628:
@stefcameron 
Just a thought, but could it be because it's running babel **after** terser minifies it? I'd be curious if that could cause babel to not transpile it.

stefcameron:
@bparish628 Thanks for the suggestion. Could be plugin order. I tried running the babel plugin as the first one in the plugin list, didn't change a thing. I can try running it after terser() to see.

[...]

PR #173, opened 2020-10-05
Transpile non-minified bundles

stefcameron:
Fixes issue #172

Also targets browsers that support ESM when transpiling for the
ESM build, rather than targetting browsers with ES5 support.

###  Features and Bug Fixes

- [x] ~Test coverage added/updated.~
- [x] ~Typings added/updated.~
- [x] ~README updated (API changes, instructions, etc.).~
- [x] ~Changes to dependencies explained.~


bparish628 (approved):
This looks like it should fix the transpiling issue we were seeing in ie11!

stefcameron:
> This looks like it should fix the transpiling issue we were seeing in ie11!

👍  Thanks for having a look. I will merge it, then, and publish, after I merge and publish https://github.com/focus-trap/tabbable/pull/97

khamiltonuk:
Hiya @stefcameron stefcameron 

Thank you for your open source work, I normally don't comment on such issues but 

I am not sure this is fully fixed in my `5.1.1` build
`node_modules/tabbable/dist/index.esm.min.js`
I can see `.map(e=>e.node).concat` this arrow function is causing a syntax error in my application in ie11 
Have I done something wrong?

stefcameron:
@khamiltonuk My pleasure! 😄 

I'm wondering why are you using the ESM build for IE11? IE11 doesn't support ESM. I would expect you would use either the CJS (`./dist/index[.min].js`) or UMD builds (`./dist/index.umd[.min].js`).

The `./dist/index.esm[.min].js` files are targeted to browsers that support ESM (which implies ES6 support, at least, which means arrow functions should be supported in those browsers also, which is probably why they didn't get transpiled to a regular `function () {}`): https://caniuse.com/?search=es%20modules

stefcameron:
I would expect your bundler to pull-in the right build of both `focus-trap` and `tabbable`.

Both project's `package.json` files are correctly configured:

Tabbable
```
  "main": "dist/index.js",
  "module": "dist/index.esm.js",
```

Focus-trap:
```
  "main": "dist/focus-trap.js",
  "module": "dist/focus-trap.esm.js",
```

jrparish:
The problem is its the bundler doing the bundling and not the browser. 
Webpack for example looks at the "module" entry first and will pick up the esm build regardless of transpilation target. 
Changing this behavior on a large project could have unintended consequences.

stefcameron:
Webpack can be configured to work around that, though. The issue is you're wanting to use ESM as a modern module system, but still expecting to have old browser-compatible code inside those modules, and not wanting to transpile any third-party modules. Is that correct (just trying to understand the needs)?

It feels strange to have diverging targets in the same bundle: ES modules (which is technically ES6) and ES5 JS. But I admit I see the same issue in my bundles that reference `focus-trap` if I set my Babel target to `ie: "11"` (I end-up with the same issue with the arrow function @khamiltonuk pointed out earlier).

Looks like this is a problem for a lot of libraries, and Webpack is keeping its position on prioritizing `module` over `main` for 3 years now: https://github.com/webpack/webpack/issues/5756 (still open)

A lot of packages seem to have solved this by adding a `browser` entry to their `package.json` (because Webpack's [default order](https://webpack.js.org/configuration/resolve/#resolvemainfields) is `mainFields: ['browser', 'module', 'main']`), which is something we could do with `focus-trap` and `tabbable`.

Apparently, [this is the browser field spec](https://github.com/defunctzombie/package-browser-field-spec), but it doesn't really get into specifics about the nature of the referenced files.

Still, given its name "browser", it feels like it should point to the UMD build.

Thoughts?

stefcameron:
I just tried this locally with tabbable's package.json having "browser: dist/index.umd.js" (with Webpack) and the resulting bundle that includes tabbable look nice. Webpack took the right UMD "path" for CJS and plopped the result right into the bundle:
```javascript
}, function(module, exports, __webpack_require__) {
  !function(exports) {
    "use strict";
    var candidateSelectors = [ "input", "select", "textarea", "a[href]", "button", "[tabindex]", "audio[controls]", "video[controls]", '[contenteditable]:not([contenteditable="false"])', "details>summary" ], candidateSelector = candidateSelectors.join(","), matches = "undefined" == typeof Element ? function() {} : Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
    function tabbable(el, options) {
      var regularTabbables = [], orderedTabbables = [];
      return getCandidates(el, (options = options || {}).includeContainer, isNodeMatchingSelectorTabbable).forEach((function(candidate, i) {
        var candidateTabindex = getTabindex(candidate);
        0 === candidateTabindex ? regularTabbables.push(candidate) : orderedTabbables.push({
          documentOrder: i,
          tabIndex: candidateTabindex,
          node: candidate
        });
      })), orderedTabbables.sort(sortOrderedTabbables).map((function(a) {           // <- NO ARROW!
        return a.node;
      })).concat(regularTabbables);
    }

    ...

    exports.focusable = focusable, exports.isFocusable = isFocusable, exports.isTabbable = isTabbable, 
    exports.tabbable = tabbable, Object.defineProperty(exports, "__esModule", {
      value: !0
    });
  }(exports);
```

stefcameron:
FYI, this conversation has now moved to https://github.com/focus-trap/tabbable/issues/99

PR #185, opened 2020-10-11
All cypress tests

liunate:
Refer back to #143 

### Done in this PR
1. Complete all cypress tests(except for one possible defect) against current demo pages. 
1. Extract common focus free-from-trap verify function for almost each test cases.

- [x] Changes to dependencies explained.
- [x] ~~README instructions updated.~~
- [x] A Changeset is __not__ added.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,7 +9,8 @@
   },
   "extends": [
     "eslint:recommended",
-    "prettier" // ALWAYS LAST: disable style rules that conflict with prettier
+    "prettier", // ALWAYS LAST: disable style rules that conflict with prettier
+    "plugin:cypress/recommended"
   ],
   "env": {
     "es6": true,
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,8 +9,8 @@
   },
   "extends": [
     "eslint:recommended",
-    "prettier", // ALWAYS LAST: disable style rules that conflict with prettier
-    "plugin:cypress/recommended"
+    "plugin:cypress/recommended",
+    "prettier" // ALWAYS LAST: disable style rules that conflict with prettier
   ],
   "env": {
     "es6": true,
--- a/cypress/integration/focus-trap-demo.spec.js
+++ b/cypress/integration/focus-trap-demo.spec.js
@@ -562,7 +562,7 @@ describe('focus-trap', () => {
         );
       });
 
-      it('click on outside element should work while keeping focus trapped inside of trap(createOptions.allowOutsideClick: function(returns true)', () => {
+      it('click on outside element should work while keeping focus trapped inside of trap(createOptions.allowOutsideClick = () => true)', () => {
         cy.get('#demo-allowoutsideclick').as('testRoot');
         cy.get('@testRoot')
           .findByRole('combobox', { name: 'Set allowOutsideClick as:' })
--- a/cypress/integration/focus-trap-demo.spec.js
+++ b/cypress/integration/focus-trap-demo.spec.js
@@ -17,6 +17,25 @@ describe('focus-trap', () => {
     cy.get(focusedElSelectorOrAliasInTrap).should('be.focused');
   }
 
+  /**
+   * Verify focus trap is **NOT** trapping focus by tab
+   * @param cyWrapper Cypress object of outside focused element yielded from `cy.get/contains/findByRole...etc`
+   */
+  function verifyFocusIsNotTrapped(cyWrapper) {
+    const outsideFocusedElAlias = `outsideFocusedEl`;
+    cyWrapper.as(outsideFocusedElAlias).should('be.focused');
+    // focus can be transitioned freely when trap is unmounted
+    let previousFocusedEl;
+    cy.get(`@${outsideFocusedElAlias}`)
+      .then(([lastlyFocusedEl]) => {
+        return (previousFocusedEl = lastlyFocusedEl);
+      })
+      .tab();
+    cy.focused().should(([nextFocusedEl]) =>
+      expect(nextFocusedEl).not.equal(previousFocusedEl)
+    );
+  }
+
   describe('demo: default', () => {
     it('traps focus tab sequence and allows deactivation by clicking deactivate button', () => {
       cy.get('#demo-default').as('testRoot');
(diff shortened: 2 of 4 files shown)


cypress/integration/focus-trap-demo.spec.js
```diff
@@ -381,22 +423,98 @@ describe('focus-trap', () => {
...
+        .should('be.focused');
+      verifyCrucialFocusTrapOnClicking('@firstElInTrap');
+
+      /*
+       focus can move into/out of iframe in trap
+       Due to cypress limitation(see below), `tab` is not tested
+      */
+
+      // Cypress limitation: `cy.tab()` skips focus on <iframe>
+      // cy.tab();
```
liunate, line 495: Cypress just does not move focus into iframe. Instead in this test trying my best to ensure focus is trapped in the `iframe` by clicking on button.

cypress/integration/focus-trap-demo.spec.js
```diff
@@ -381,22 +423,98 @@ describe('focus-trap', () => {
...
+      cy.get('@testRoot').find('iframe').as('iframeInTrap');
+      cy.get('@iframeInTrap').then(([iframe]) => {
+        // Cypress limitation: TypeError: Cannot read property 'call' of undefined
+        // cy.wrap(iframe.contentDocument).find('a').should('exist');
+        const buttonEl = iframe.contentDocument.querySelector('button');
+        cy.wrap(buttonEl).as('buttonInIFrame').click();
+        cy.get('@buttonInIFrame').should('be.focused');
+
+        verifyCrucialFocusTrapOnClicking('@buttonInIFrame');
+      });
```
liunate, line 505: There is a npm package `cypress-iframe` but it seems to not work fine that its `cy.iframe().find...` just does not work. Anyway, trying my best to ensure focus is trapped in iframe.

cypress/integration/focus-trap-demo.spec.js
```diff
@@ -381,22 +423,98 @@ describe('focus-trap', () => {
...
+        .findByRole('button', { name: 'deactivate trap' })
+        .as('deactivateElInTrap')
+        .click();
+      verifyFocusIsNotTrapped(cy.get('@lastlyFocusedElBeforeTrapIsActivated'));
+    });
   });
 
   describe('demo: allowoutsideclick', () => {
-    // TODO
+    // TODO: defect? this only traps keyboard but click on outside element is still changing document.activeElement
```
liunate, line 517: This is the only left test case I am not sure about. The keyboard `tab` is trapped when it's activated but click on the demo page is not trapped at all. Will come back to this one in the next few days.

.eslintrc.json
```diff
@@ -9,7 +9,8 @@
   },
   "extends": [
     "eslint:recommended",
-    "prettier" // ALWAYS LAST: disable style rules that conflict with prettier
+    "prettier", // ALWAYS LAST: disable style rules that conflict with prettier
+    "plugin:cypress/recommended"
```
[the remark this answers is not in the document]
stefcameron: Sounds like a good addition! But now it's _last_ in the last, after "prettier" where we have a comment stating it should always be last... 😉  I wrote that comment, but in retrospect, it's probably always supposed to be _after_ "eslint:recommended" since I think Prettier's rules probably only know about basic/built-in ESLint rules. They can't be monitoring every ESLint plugin in existence...

So would you mind updating the "ALWAYS LAST" statement to "AFTER ESLINT" as part of this change?

cypress/integration/focus-trap-demo.spec.js
stefcameron: Thanks for doing some research! I'd rather skip something than use a potentially unstable/non-working plugin that could make the test flaky.

cypress/integration/focus-trap-demo.spec.js
stefcameron: This trap allows clicking outside of it while maintaining an active trap, if I remember correctly. That would explain why our "verify focus is trapped" utility function doesn't work, because the utility function expects that clicking on those repo links will do nothing -- but they end-up trying to load the page, and I'll bet you were seeing the test fail with an error from Cypress talking about some cross-domain error thingamagig.

I ran into the same thing when I recently added another demo and test case [clickoutsidedeactivates](https://github.com/focus-trap/focus-trap/blob/master/cypress/integration/focus-trap-demo.spec.js#L402) which is closely related to the pre-existing "allowoutsideclick" demo. I wrote a test-specific [checkTrap()](https://github.com/focus-trap/focus-trap/blob/master/cypress/integration/focus-trap-demo.spec.js#L410) helper because, for this test, since it also allows outside clicks, we can only verify that focus is trapped by running the expected tab sequence back and forth...

cypress/integration/focus-trap-demo.spec.js
```diff
@@ -17,6 +17,25 @@ describe('focus-trap', () => {
     cy.get(focusedElSelectorOrAliasInTrap).should('be.focused');
   }
 
+  /**
+   * Verify focus trap is **NOT** trapping focus by tab
+   * @param cyWrapper Cypress object of outside focused element yielded from `cy.get/contains/findByRole...etc`
+   */
+  function verifyFocusIsNotTrapped(cyWrapper) {
+    const outsideFocusedElAlias = `outsideFocusedEl`;
```
stefcameron, line 25: 🤔  nit: I'm surprised ESLint didn't complain about backticks for a string that doesn't use variable references. Must be a rule I expect is "recommended" but isn't.

stefcameron (changes requested):
This all looks great. A BIG THANK YOU for pressing on to add all but one remaining test! It's a huge help to me and to the consumers of this library. Just one thing to change regarding the Cypress ESLint plugin addition (see related comment).

.eslintrc.json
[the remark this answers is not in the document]
liunate: That's a fair point, though I think **AFTER ESLINT** feels somehow unclear since every plugin is for eslint.

I am just wondering if it's still better to keep `prettier` at last, so it is always the bottom line of coding style regardless whatever plugin inserted before `prettier`. In this case I can move `prettier` to last, how do you think?

cypress/integration/focus-trap-demo.spec.js
liunate: Ooh okay now I get the idea. I will add one test for this.

.eslintrc.json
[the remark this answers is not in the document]
liunate: > That's a fair point, though I think **AFTER ESLINT** feels somehow unclear since every plugin is for eslint.
> 
> I am just wondering if it's still better to keep `prettier` at last, so it is always the bottom line of coding style regardless whatever plugin inserted before `prettier`. In this case I can move `prettier` to last, how do you think?

I just changed `prettier` to last, see if you think the other way around. I am okay to change it back. 🥂

Also I added the only missing `allowClickOutside` test now with all tests in place I should feel safer to help review and enhance the library 🎉

liunate:
Most change to existing `describe('demo: clickoutsidedeactivates'` are only changing the indent. It's because the last added test `describe('demo: allowoutsideclick')` piggybacking onto the `activateTrap` and `checkTrap` <--- saved me lots of time tks 👏

.eslintrc.json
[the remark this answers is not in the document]
stefcameron: Sounds good on both points, thanks!

stefcameron:
Resolved conflict in package.json

stefcameron:
Something strange is up with the e2e action when it tries to "restore cached NPM modules". This hasn't happened in the past, not sure what the issue is. I'll wait a few hours and try again.

stefcameron (approved):
This looks good to me now, and the e2e tests passed on CI, finally (some GH issue was fixed by someone somewhere)! Thanks again for finishing-off all the tests! 🎉

liunate:
Thank you for the review😊

Issue #199, opened 2020-10-22
Feature request: allowOutsideClick for keyboard events


zioth:
`allowOutsideClick` can be used to allow certain page elements to be clickable outside the focus trap. However, these elements do not receive keyboard events, including tabs, which is an accessibility issue. My feature request is either to pass keyboard events and `tabbable` filtering through `allowOutsideClick`, or else create an `allowOutsideKeystrokes` option.

stefcameron:
Thanks for your thoughts on this option. The goal of this library is to trap focus within a certain "boundary" or "scope" element. The mouse (and our fingers, in the case of touch) has an advantage over the keyboard in that it's possible to click/touch an element __without__ letting focus change to that element, hence the name of this option to allow outside _clicks_.

Unfortunately for the keyboard, I don't believe there's a way to direct keystrokes to an element _without_ first setting focus to that element. Adding an `allowOutsideKeystrokes` option sounds like it would violate the very nature of a "focus trap".

The point is to trap focus inside a container element, not letting it get outside of it. If you need keyboard focus to get to some other element outside the trap, perhaps there's a design flaw that needs to be fixed?

zioth:
What you're saying makes perfect sense from a philosophical perspective - if a focus trap doesn't always trap focus, then it's not what it says it is. However, there are cases where you want a specific element to be interactable, even though it's outside the trap.

Here's an example: I open a dialog, and trap focus to that dialog. While the dialog is still open, the client's messaging system (which is not a child node of the dialog) pops up an error. The user cannot dismiss the error, unless you use the function version of `allowOutsideClick` to create an exception, but even with that, keyboard users still can't dismiss it. This creates both an accessibility issue, and some UX confusion, because `:active` and `:focus` states are never set.

Rather than my initial `allowOutsideKeystrokes` suggestion, which is a bit awkward, what about an `isException` callback, which behaves like this?
```
isException: e => e.target.id = 'error_message'
```
It could be called in every click and keyboard event, as well as being used to calculate the list of tabbable elements.

stefcameron:
Thank you for that example: It illustrates the need nicely!

We could explore a solution that uses a callback to check if a node should be an exception, as you proposed.

Here's what I'm thinking:

[...]

PR #217, opened 2020-11-13
Add support for passing in multiple elements

cgood92:
Add support for passing in multiple elements

Note that only passing in a single element is still supported

###  Features and Bug Fixes

- [X] Test coverage added/updated.
- [x] Typings added/updated.

-------------------
Demo 1
demo2

Demo 2
demo

### DIFF
--- a/.changeset/tidy-suns-notice.md
+++ b/.changeset/tidy-suns-notice.md
@@ -1,0 +1,5 @@
+---
+'focus-trap': patch
+---
+
+Adding support for multiple elements to be passed in
--- a/.changeset/tidy-suns-notice.md
+++ b/.changeset/tidy-suns-notice.md
@@ -1,5 +1,5 @@
 ---
-'focus-trap': patch
+'focus-trap': minor
 ---
 
-Adding support for multiple elements to be passed in
+Adding support for multiple elements to be passed in #217
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Returns a new focus trap on `element`.
 
 `element` can be
 - a DOM node (the focus trap itself) or
-- a selector string (which will be pass to `document.querySelector()` to find the DOM node) or
+- a selector string (which will be passed to `document.querySelector()` to find the DOM node) or
 - an array of DOM nodes or selector strings (where the order determines where the focus will go after the last tabbable element of a DOM node/selector is reached).
 
 `createOptions`:
@@ -132,6 +132,16 @@ Unpause an active focus trap. (See `pause()`, above.)
 Focus is forced into the trap just as described for `focusTrap.activate()`.
 
 If the focus trap has not been activated or has not been paused, nothing happens.
+
+Returns the `focusTrap`.
+
+### focusTrap.updateContainerElements()
+
+Update the element(s) that are used as containers for the focus trap.
+
+When you call the function `createFocusTrap`, you pass in an element (or selector), or an array of elements (or selectors) to keep the focus within.  This method simply allows you to update which elements to keep the focus within.
+
+A use case for this is found in focus-trap-react, where React `ref`'s may not be initialized yet, but when they are you want to have them be a container element.
 
 Returns the `focusTrap`.
 
--- a/README.md
+++ b/README.md
@@ -63,7 +63,8 @@ Returns a new focus trap on `element`.
 
 `element` can be
 - a DOM node (the focus trap itself) or
-- a selector string (which will be pass to `document.querySelector()` to find the DOM node).
+- a selector string (which will be pass to `document.querySelector()` to find the DOM node) or
+- an array of DOM nodes or selector strings (where the order determines where the focus will go after the last tabbable element of a DOM node/selector is reached).
 
 `createOptions`:
 
--- a/cypress/integration/focus-trap-demo.spec.js
+++ b/cypress/integration/focus-trap-demo.spec.js
@@ -774,6 +774,85 @@ describe('focus-trap', () => {
       cy.get('@testRoot')
         .findByRole('button', { name: 'deactivate trap' })
         .click();
+
+      // focus can be transitioned freely when trap is deactivated
+      cy.findByRole('heading', { name: 'focus-trap demo' })
+        .as('outsideEl')
+        .click();
+      verifyFocusIsNotTrapped(cy.get('@outsideEl'));
+    });
+  });
+
+  describe('demo: multiple traps with multiple elements', () => {
+    it('multiple traps with multiple elements works', () => {
+      cy.get('#demo-multipleelements-multipletraps').as('testRoot');
+
+      // activate trap 1
+      cy.get('@testRoot')
+        .findByRole('button', { name: 'activate trap 1' })
+        .as('lastlyFocusedElementBeforeTrapIsActivated')
+        .click();
+
+      // Focus should be in trap 1
+      cy.get('@testRoot')
+        .findByRole('link', { name: 'with' })
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'some')
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'focusable')
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'See')
+        .should('be.focused');
+
+      // activate focus trap 2.  This should pause trap 1
+      cy.get('@testRoot')
+        .findByRole('button', { name: 'activate trap 2' })
+        .click();
+
+      // Focus should be in trap 2
+      cy.get('@testRoot')
+        .findByRole('link', { name: 'something' })
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'last')
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'area')
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'something')
+        .should('be.focused');
+
+      // stop focus trap 2
+      cy.get('@testRoot')
+        .findByRole('button', { name: 'deactivate trap 2' })
+        .click();
+
+      // focus should resume back to trap 1
+      cy.get('@testRoot')
+        .findByRole('link', { name: 'See' })
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'how')
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'works')
+        .should('be.focused');
+
+      // focus can be transitioned freely when both traps are deactivated
+      cy.get('@testRoot')
+        .findByRole('button', { name: 'deactivate trap 1' })
+        .as('lastButtonClicked')
+        .click();
+
+      // focus can be transitioned freely when trap is deactivated
+      cy.findByRole('heading', { name: 'focus-trap demo' })
+        .as('outsideEl')
+        .click();
+      verifyFocusIsNotTrapped(cy.get('@outsideEl'));
     });
   });
 });
--- a/cypress/integration/focus-trap-demo.spec.js
+++ b/cypress/integration/focus-trap-demo.spec.js
@@ -711,4 +711,69 @@ describe('focus-trap', () => {
       verifyFocusIsNotTrapped(cy.get('@outsideFocusedEl'));
     });
   });
+
+  describe('demo: multiple elements passed in', () => {
+    it('can accept multiple elements passed in, and keep the focus within the elements', () => {
+      cy.get('#demo-multipleelements').as('testRoot');
+
+      // activate trap
+      cy.get('@testRoot')
+        .findByRole('button', { name: 'activate trap' })
+        .as('lastlyFocusedElementBeforeTrapIsActivated')
+        .click();
+
+      // 1st element should be focused
+      cy.get('@testRoot')
+        .findByRole('link', { name: 'with' })
+        .as('firstElementInTrap')
+        .should('be.focused');
+
+      // trap is active(keep focus in trap by tabbing through the focus trap's tabbable elements)
+      cy.get('@firstElementInTrap')
+        .tab()
+        .should('have.text', 'some')
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'focusable')
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'See')
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'how')
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'works')
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'with')
+        .should('be.focused')
+        .tab({ shift: true })
+        .should('have.text', 'works')
+        .should('be.focused')
+        .tab({ shift: true })
+        .should('have.text', 'how')
+        .should('be.focused')
+        .tab({ shift: true })
+        .should('have.text', 'See')
+        .should('be.focused')
+        .tab({ shift: true })
+        .should('have.text', 'focusable')
+        .should('be.focused')
+        .tab({ shift: true })
+        .should('have.text', 'some')
+        .should('be.focused')
+        .tab({ shift: true })
+        .should('have.text', 'with')
+        .should('be.focused')
+        .tab({ shift: true })
+        .should('have.text', 'works')
+        .should('be.focused');
+
+      // focus can be transitioned freely when trap is deactivated
+      cy.get('@testRoot')
+        .findByRole('button', { name: 'deactivate trap' })
+        .click();
+    });
+  });
 });
--- a/demo/index.html
+++ b/demo/index.html
@@ -166,7 +166,7 @@
           activate inner trap
         </button>
       </p>
-      <div id="nested-nested" style="display:none;padding:5px 10px;background:pink;">
+      <div id="nested-nested" style="display:none;padding:5px 10px;">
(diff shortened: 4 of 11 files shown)


stefcameron:
@cgood92 Wow, this is interesting! Thank you. I haven't reviewed yet, but my first thought is I wonder if this would fix #199. WDYT?

stefcameron:
@zioth WDYT about this? Do you think it could be used to solve your use case in #199?

zioth:
> @zioth WDYT about this? Do you think it could be used to solve your use case in #199?

I think so!

cgood92:
After this PR is merged, this PR will be good to go: https://github.com/focus-trap/focus-trap-react/pull/179, after a version bump of course.

index.js
```diff
@@ -296,21 +312,64 @@ function createFocusTrap(element, userOptions) {
   // kind of need to capture the action at the keydown phase.
   function checkTab(e) {
     updateTabbableNodes();
-    if (e.shiftKey && e.target === state.firstTabbableNode) {
-      e.preventDefault();
-      tryFocus(state.lastTabbableNode);
-      return;
+    const isShift = e.shiftKey;
+
+    if (isShift) {
```
stefcameron, line 317: Could we collapse this into `if (e.shiftKey)`?

index.js
```diff
@@ -296,21 +312,64 @@ function createFocusTrap(element, userOptions) {
...
+        tryFocus(destinationNode);
+
+        return;
+      }
     }
-    if (!e.shiftKey && e.target === state.lastTabbableNode) {
-      e.preventDefault();
-      tryFocus(state.firstTabbableNode);
-      return;
+    if (!isShift) {
```
stefcameron, line 342: Can this be an `else` statement, no need for `return` statements on 339 and 365...?

index.js
```diff
@@ -296,21 +312,64 @@ function createFocusTrap(element, userOptions) {
...
-      return;
+    const isShift = e.shiftKey;
+
+    if (isShift) {
+      const startOfGroupIndex = state.tabbableGroups.findIndex(
+        ({ firstTabbableNode }) => e.target === firstTabbableNode
+      );
+
+      const isStartOfAGroup = startOfGroupIndex !== -1;
+      if (isStartOfAGroup) {
```
stefcameron, line 323: Could we check for `>= 0` instead of strictly looking for a single value? I prefer not to depend on a single value whenever possible. "If the index is any negative value, then it's not the start of a group..."

Also, I don't think we ever use the `isStartOfAGroup` variable anywhere, so why not just `if (startOfGroupIndex >= 0)`?

index.js
```diff
@@ -296,21 +312,64 @@ function createFocusTrap(element, userOptions) {
...
-      e.preventDefault();
-      tryFocus(state.firstTabbableNode);
-      return;
+    if (!isShift) {
+      const lastOfGroupIndex = state.tabbableGroups.findIndex(
+        ({ lastTabbableNode }) => e.target === lastTabbableNode
+      );
+
+      const isLastOfAGroup = lastOfGroupIndex !== -1;
+      if (isLastOfAGroup) {
```
stefcameron, line 348: Similarly to the shift case above, could this be simplified to `if (lastOfGroupIndex >= 0)`?

index.js
```diff
@@ -296,21 +312,64 @@ function createFocusTrap(element, userOptions) {
...
+        ({ firstTabbableNode }) => e.target === firstTabbableNode
+      );
+
+      const isStartOfAGroup = startOfGroupIndex !== -1;
+      if (isStartOfAGroup) {
+        e.preventDefault();
+
+        const isFirstGroup = startOfGroupIndex === 0;
+        const lastGroupIndex = state.tabbableGroups.length - 1;
+        const previousGroupIndex = startOfGroupIndex - 1;
```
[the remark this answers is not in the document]
stefcameron: So maybe just this? We're just needing to go backward one, and loop around if we're already at the start.
```javascript
const destinationGroupIndex = startOfGroupIndex === 0
  ? state.tabbableGroups.length - 1
  : startOfGroupIndex - 1;
const destinationGroup = state.tabbableGroups[destinationGroupIndex]; // we know we'll get a group
const destinationNode = destinationGroup.lastTabbableNode;
```

index.js
```diff
@@ -296,21 +312,64 @@ function createFocusTrap(element, userOptions) {
...
+          lastOfGroupIndex === state.tabbableGroups.length - 1;
+        const firstGroupIndex = 0;
+        const nextGroupIndex = lastOfGroupIndex + 1;
+        const destinationGroupIndex = isLastGroup
+          ? firstGroupIndex
+          : nextGroupIndex;
+
+        const destinationGroup = state.tabbableGroups[destinationGroupIndex];
+        const destinationNode =
+          destinationGroup && destinationGroup.firstTabbableNode;
```
stefcameron, line 361: Similarly to the shift case, I think we can simplify this down to 3 statements and 3 variables.

In fact, we could also de-dup the `tryFocus(destinationNode)` statement by declaring `let destinationNode` before the IF statement, and then call `tryFocus(destinationNode)` after the IF statement, as the last thing the function does before returning.

demo/index.html
```diff
@@ -459,6 +459,36 @@ <h2 id="setreturnfocus-heading">setReturnFocus option</h2>
     </div>
   </div>
 
+  <div id="demo-multipleelements">
```
stefcameron, line 462: Thank you for adding a test case. This is great. I feel we need another one that tests this multi-container feature along with multiple traps where we're able to activate a second trap, having the first one "paused", and each trap is able to deal with its own set of multiple containers, just to flex that muscle a bit and make sure it all works as expected.

cypress/integration/focus-trap-demo.spec.js
```diff
@@ -711,4 +711,69 @@ describe('focus-trap', () => {
...
+        .should('be.focused')
+        .tab({ shift: true })
+        .should('have.text', 'works')
+        .should('be.focused');
+
+      // focus can be transitioned freely when trap is deactivated
+      cy.get('@testRoot')
+        .findByRole('button', { name: 'deactivate trap' })
+        .click();
+    });
```
stefcameron, line 777: Please add a `verifyFocusIsNotTrapped(cy.get('@outsideFocusedEl'));` check at the very end of the test to verify focus has been freed. We should be doing this for all tests.

README.md
```diff
@@ -63,7 +63,8 @@ Returns a new focus trap on `element`.
 
 `element` can be
 - a DOM node (the focus trap itself) or
-- a selector string (which will be pass to `document.querySelector()` to find the DOM node).
+- a selector string (which will be pass to `document.querySelector()` to find the DOM node) or
+- an array of DOM nodes or selector strings (where the order determines where the focus will go after the last tabbable element of a DOM node/selector is reached).
```
stefcameron, line 67: Please also add into in the README about the new `updateContainerElements()` method on the trap object returned from `createFocusTrap()`, further down, where we detail each method, e.g. at line 137.

stefcameron (changes requested):
@cgood92 Nice work! My comments are mostly geared to crossing some Ts and dotting some Is, and maybe simplifying a bit of the code, but overall, I think the change is pretty solid!

Now I'm thinking of #199 in context of this change. I think that for the use case [described](https://github.com/focus-trap/focus-trap/issues/199#issuecomment-718304363) there about the notification while in a modal, we would need some way that an __existing trap__ can be given a new set of container elements, possibly entirely different from that with which it was initial created.

I believe that's what you've enabled with the new `updateContainerElements()` method on the trap object. Is that correct?

index.d.ts
```diff
@@ -101,7 +101,7 @@ declare module 'focus-trap' {
    *  find the element.
    */
   export function createFocusTrap(
-    element: HTMLElement | string,
+    element: HTMLElement | string | Array<HTMLElement | string>,
```
stefcameron, line 104: Please also add the new `updateContainerElements()` method to the `FocusTrap` interface.

cgood92:
@stefcameron thank you very much for the timely review.  I appreciate the comments.  I have gone through and made every adjustment asked for.  

If it is all right, not at all to be contrary or rude, I may follow up on a few of the comments, just to shed a different perspective - all for the sake of introducing to new ideas, not at all to banter or anything.  Again, I've already made the changes, so I'm not arguing for their need or not, just giving some different opinions, that is all.

Let me know if there is anything else I can do!  At Workfront, we are using this component, and would love to update as soon as we can.

.changeset/tidy-suns-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+'focus-trap': patch
```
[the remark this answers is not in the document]
cgood92: I thought of this later as well, you are of course right, this is a minor.  Good call.

.changeset/tidy-suns-notice.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Adding support for multiple elements to be passed in
```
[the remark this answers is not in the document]
cgood92: Brilliant.  Done.

README.md
```diff
@@ -63,7 +63,8 @@ Returns a new focus trap on `element`.
 
 `element` can be
 - a DOM node (the focus trap itself) or
-- a selector string (which will be pass to `document.querySelector()` to find the DOM node).
+- a selector string (which will be pass to `document.querySelector()` to find the DOM node) or
```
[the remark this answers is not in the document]
cgood92: Thanks :)

index.js
```diff
@@ -62,10 +60,25 @@ function createFocusTrap(element, userOptions) {
...
     pause: pause,
     unpause: unpause,
+    updateContainerElements: updateContainerElements,
   };
 
+  updateContainerElements(elements);
+
   return trap;
 
+  function updateContainerElements(elements) {
```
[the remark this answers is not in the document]
cgood92: I like calling the arguments and the parameters by the same name.  I also see value in the shadowing principle you're referring to.  Done.

index.js
```diff
@@ -49,8 +47,8 @@ function createFocusTrap(element, userOptions) {
   };
 
   var state = {
-    firstTabbableNode: null,
-    lastTabbableNode: null,
+    containers: [],
+    tabbableGroups: [],
```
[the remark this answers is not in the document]
cgood92: Good call, done.

index.js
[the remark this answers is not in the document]
cgood92: You were right about not needing to check if we have the group or not.  Thanks for that.

I don't necessarily think that consolidating to fewer lines is always beneficial.  I like naming logic with variables, so that I can read the code like in english in my head.  I don't think your request is at all on another extreme, but I'm just explaining that it was intentional to have good variable names and make each statement simple to read and understand.

I made the changes however.

index.js
cgood92: Since we're moving to the `else` below, this makes even more sense.

index.js
cgood92: Done

demo/index.html
cgood92: Done.

cypress/integration/focus-trap-demo.spec.js
cgood92: Done (I think correctly)

index.d.ts
cgood92: Good call.  Done.

README.md
cgood92: Done.

index.js
cgood92: Done.  However, just as a different thought, I actually prefer the `-1` check here.

`-1` has a special meaning here.  It means that it's not found in the array.  When you check against `-1`, it's clear (in my mind) that what you're doing is checking to see if it is part of the array or not.

When you use `>=0`, it takes me a minute longer to see what is going on here.  The `>=0` makes it seem like I'm checking against a range, where the number may be `-2`, `-1`, `-13`, or some positive number.  When really, I'm not checking for a range condition, the only possible negative number here is -1.

I also like naming it, for reasons I stated above.  I guess I took that from Clean Code by Martin.

I only add this as a varying opinion, not at all to banter.  Just spreading ideas.

index.js
cgood92: Done with same comments.

index.js
cgood92: Done.

stefcameron:
@cgood92 
> @stefcameron thank you very much for the timely review. I appreciate the comments. I have gone through and made every adjustment asked for.

Thank you! 😄  Checking this out... 👀 

> If it is all right, not at all to be contrary or rude, I may follow up on a few of the comments, just to shed a different perspective - all for the sake of introducing to new ideas, not at all to banter or anything. Again, I've already made the changes, so I'm not arguing for their need or not, just giving some different opinions, that is all.

Certainly, I appreciate the different perspective!

> Let me know if there is anything else I can do! At Workfront, we are using this component, and would love to update as soon as we can.

Cool! Hey, Adobe is acquiring Workfront, nice. Well, assuming you like Adobe. I worked for Adobe for 10 years. Great company!

index.d.ts
```diff
@@ -91,6 +91,7 @@ declare module 'focus-trap' {
     deactivate(deactivateOptions?: DeactivateOptions): void;
     pause(): void;
     unpause(): void;
+    updateContainerElements(containerElements: HTMLElement | string | Array<HTMLElement | string>): void;
```
stefcameron, line 94: Since the method returns the trap, return type should be `FocusTrap` instead of `void`. But then, none of the other existing API methods return the trap. I would say it's nice to enable chaining on these methods by returning the trap, including this new method, but I'm thinking it might actually be better _not_ to return the trap to keep things consistent, unless that's critical for something I'm not thinking of. All methods could be updated to return the trap in a separate PR. WDYT?

README.md
```diff
@@ -134,6 +135,16 @@ If the focus trap has not been activated or has not been paused, nothing happens
...
 
+### focusTrap.updateContainerElements()
+
+Update the element(s) that are used as containers for the focus trap.
+
+When you call the function `createFocusTrap`, you pass in an element (or selector), or an array of elements (or selectors) to keep the focus within.  This method simply allows you to update which elements to keep the focus within.
+
+A use case for this is found in focus-trap-react, where React `ref`'s may not be initialized yet, but when they are you want to have them be a container element.
+
+Returns the `focusTrap`.
```
stefcameron, line 146: Marking this here so I don't forget as something to remove depending on your thoughts on my comment in `index.d.ts`...

index.js
```diff
@@ -62,10 +62,27 @@ function createFocusTrap(element, userOptions) {
...
+
+    state.containers = elementsAsArray.map((element) =>
+      typeof element === 'string' ? doc.querySelector(element) : element
+    );
+
+    if (state.active) {
+      updateTabbableNodes();
+    }
+
+    return trap;
```
stefcameron, line 83: Marking this here (so I don't forget) as something to remove depending on your thoughts on my comment in `index.d.ts`...

stefcameron (commented):
Thanks for adding that additional test case, as well as providing your perspective on use of multiple variables. I can appreciate how it can help with readability.

One comment I'd like your thoughts on...

index.d.ts
cgood92: They actually all return the trap.  They all say that in the README.md.  So the inaccurate thing here is the TS return type.

I put void cause that's what the others are doing.  But, they all should return a trap.

I don't know TS that well, so I didn't update the others.  But, now I see you already have a FocusTrap TS type, so I can actually easily change this and the others.  Let me do that real quick.

cgood92: √ Updated

stefcameron: Thanks! Ah, you're right about `activate()` and `deactivate()`, but I don't see that happening for [pause and unpause](https://github.com/focus-trap/focus-trap/blob/master/index.js#L122-L133) as it is.

Do you mind adding that? Note there are 2 return paths in each method (the way they're currently implemented).

Then everything will be synced-up. 😄

stefcameron:
@cgood92 I saw an issue with styling in the multi-trap demo, and also thought we could enhance the demos to show off this functionality a bit better with some highlights, so I just pushed https://github.com/focus-trap/focus-trap/pull/217/commits/4bc72f0af01fd18f63c69b9025df310f1a1e4444 to your fork. We can revert if you don't like it, but here are the befores and afters:

Single trap before
multi-trap-single-before

Single trap after
multi-trap-single-after

Multi trap before
multi-traps-before

Multi trap after
multi-traps-after

No functional changes to anything, though.

stefcameron:
I have to say, this change is looking solid! I think we can declare it done after fixing the returns from `pause()` and `unpause()`. I'll do a release right after.

index.d.ts
cgood92: Done.  Thanks for checking.

cgood92:
Looks all good to go.  Thanks!  @stefcameron!

stefcameron (approved):
This looks solid to me! 💪

stefcameron:
Fixes #199

stefcameron:
all-contributors please add @cgood92 for code, docs, examples, tests

stefcameron:
@all-contributors please add @cgood92 for code, docs, examples, tests

stefcameron:
Published in 6.2.0

PR #222, opened 2020-11-17
Refactoring to use const/let, and simplify a few lines

cgood92:
###  Features and Bug Fixes

- [x] Typings added/updated.
- [x] Changes to dependencies explained.

Since I was in here, I thought I would put up a simple PR that had no functionality change.  These changes are just about changing vars to let/const (no more vars in repo), simplifying a few lines, etc.

Not offended if these are not desirable.

### DIFF
--- a/.changeset/sweet-wombats-jam.md
+++ b/.changeset/sweet-wombats-jam.md
@@ -1,0 +1,5 @@
+---
+'focus-trap': patch
+---
+
+Refactoring to use const/let, and simplify a few lines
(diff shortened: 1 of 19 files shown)


cgood92:
However, please don't get too distracted with this PR before the focus-trap-react is merged and done.  We are excited to move on with that quickly.

stefcameron (approved):
@cgood92 Thanks for taking the time to do this! I've been wanting to do the same thing for while. 😄

PR #237, opened 2020-11-25
Allow trap containers to have no tabbable nodes

### LINKED ISSUE #223: If a `containerElements` element doesn't exist, focus-trap is confused
If you set `containerElements` to `[null]`, every tabbable node in the document is reachable. I believe this also happens when one of the container elements is no longer in the DOM.

stefcameron:
Fixes #223 and adds an exception thrown if the trap is allowed to get into a state where none of its containers contain any tabbable nodes (there must always be at least one container that has at least one tabbable node among all containers configured) -- unless the `fallbackFocus` option has been configured to provide a fallback node.

- [x] Add test coverage
- [x] Add changeset

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,8 +69,16 @@ const createFocusTrap = function (elements, userOptions) {
   const state = {
     // @type {Array<HTMLElement>}
     containers: [],
-    // @type {{ firstTabbableNode: HTMLElement, lastTabbableNode: HTMLElement }}
+
+    // list of objects identifying the first and last tabbable nodes in all containers/groups in
+    //  the trap
+    // NOTE: it's possible that a group has no tabbable nodes if nodes get removed while the trap
+    //  is active, but the trap should never get to a state where there isn't at least one group
+    //  with at least one tabbable node in it (that would lead to an error condition that would
+    //  result in an error being thrown)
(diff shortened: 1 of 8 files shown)


stefcameron:
@zioth Please have a look! I tested this locally with `focus-trap-react` and your exact use case (from the sandbox you provided in #223) and the issue is fixed.

issue-223-fixed

index.js
```diff
@@ -213,10 +237,14 @@ const createFocusTrap = function (elements, userOptions) {
   const checkFocusIn = function (e) {
     // In Firefox when you Tab out of an iframe the Document is briefly focused.
     if (containersContain(e.target) || e.target instanceof Document) {
-      return;
+      if (containersContain(e.target)) {
```
cgood92, line 240: Does it feel redundant to check this twice?  `containersContain(e.target)` in the line above and `containersContain(e.target)` here again?

cgood92:
Besides the one non-important comment, this looks good to me.

cgood92 (approved):

index.js
stefcameron: Yes, very! 😄  Thanks for pointing that out, fixed.

PR #407, opened 2021-05-29
Improved support for animated dialogs

Dan503:
- Added `checkCanActivate` and `onSuccessfulActivation` settings for better animated dialog support.
- Added links to index.html to make it easier to view source code of demos

Fixes issue: https://github.com/focus-trap/focus-trap-react/issues/239

<details>
<br/>

- Typings added/updated.

</details>


index.js
```diff
@@ -3,7 +3,7 @@ var xtend = require('xtend');
 
 var listeningFocusTrap = null;
 
-function focusTrap(element, userOptions) {
+function createFocusTrap(element, userOptions) {
```
Dan503, line 6: Don't worry, this is backwards compatible since it is a default export.

I renamed it to be consistent with the name that is used in the demos to avoid confusion.

index.js
```diff
@@ -34,24 +34,68 @@ function focusTrap(element, userOptions) {
...
+        updateTabbableNodes();
+        addListeners();
+        if (onSuccessfulActivation) {
+          onSuccessfulActivation();
+        }
+      });
+    } else {
+      updateTabbableNodes();
+      if (onActivate) {
+        onActivate();
```
Dan503, line 93: I would like to call `onActivate` before `updateTabbableNodes` here to be consistent with the `checkCanActivate` version but that could potentially break existing code so it needs to come after `updateTabbableNodes()`

stefcameron:
> Renamed "demos" folder to "docs" so the project can be published to GitHub pages

Hi, @Dan503, thanks for making this PR to solve that issue over on `focus-trap-react`! I haven't really looked yet, but this caught my eye in the summary.

There shouldn't be any need to rename this directory to make it possible to use GitHub pages. We do that already! 😄 This is how it's done, quite simple: https://github.com/focus-trap/focus-trap/blob/master/CONTRIBUTING.md#updating-the-demo

What did you run into that made you think it was necessary to rename the directory? Seems to be causing a lot of conflicts...

I think I'd prefer to leave that to another PR -- if it's really necessary.

Dan503:
This is how I normally handle github pages in my projects. I make the `docs` folder the root folder for the website. GH Pages only supports root or `/docs` as the root folder for the site.

SmartSelect_20210530-090346_Chrome

SmartSelect_20210530-090411_Chrome

I get what you are doing now though, you are setting GH pages to look at the root folder and then adding `/demos` to the url. 

How is GH Pages working with the bundled js file being ignored by git though?
Git Hub pages only publishes files if they are stored in Git. Do you have a github action that compiles it for you or something?

The GH pages change isn't really necessary. I'll revert that.

Dan503:
No wonder there are merge conflicts. For some reason I was working off a branch that was running on v3 and focus-trap is up to v6.

I need to redo my PR.

index.js
```diff
@@ -470,21 +494,45 @@ const createFocusTrap = function (elements, userOptions) {
...
+      const onActivate = getOption('onActivate');
+      const onSuccessfulActivation = getOption('onSuccessfulActivation');
+      const checkCanActivate = getOption('checkCanActivate');
+
+      if (checkCanActivate) {
+        if (onActivate) {
+          onActivate();
+        }
+
+        state.containers.forEach((container) => {
```
Dan503, line 516: I'm uncertain about this.

It works for my use case but my use case only has a single focus-trap container. I'm not sure if this will work for multiple focus-traps.

package.json
```diff
@@ -23,9 +23,9 @@
     "format:watch": "onchange \"{*,src/**/*,test/**/*,demo/js/**/*,.github/workflows/*,cypress/**/*}.+(js|yml)\" -- prettier --write {{changed}}",
     "lint": "eslint \"*.js\" \"demo/**/*.js\" \"cypress/**/*.js\"",
     "clean": "rm -rf ./dist",
-    "compile:esm": "BUILD_ENV=esm BABEL_ENV=esm rollup -c",
-    "compile:cjs": "BUILD_ENV=cjs BABEL_ENV=es5 rollup -c",
-    "compile:umd": "BUILD_ENV=umd BABEL_ENV=es5 rollup -c",
+    "compile:esm": "cross-env BUILD_ENV=esm BABEL_ENV=esm rollup -c",
+    "compile:cjs": "cross-env BUILD_ENV=cjs BABEL_ENV=es5 rollup -c",
+    "compile:umd": "cross-env BUILD_ENV=umd BABEL_ENV=es5 rollup -c",
```
Dan503, line 28: I added [cross-env][1] because `BUILD_ENV=X` doesn't work on windows.

[1]: https://www.npmjs.com/package/cross-env

demo/index.html
```diff
@@ -25,7 +25,7 @@ <h1 tabindex="0">focus-trap demo</h1>
     <div id="demo-default">
       <h2 id="default-heading">default behavior</h2>
       <p>
-        <button id="activate-default" aria-label="activate trap for 'defaults' demo">
+        <button id="activate-default" title="activate trap for 'defaults' demo">
```
Dan503, line 28: I changed `aria-label` to `title` because aria-label will completely override the visible label which is an Accessibility fail against [WCAG criteria 2.5.3: Label in Name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name)

README.md
```diff
@@ -172,25 +176,16 @@ const { createFocusTrap } = require('../../dist/focus-trap');
...
 
 const focusTrap = createFocusTrap('#default', {
-  onActivate: function () {
-    container.className = 'trap is-active';
-  },
-  onDeactivate: function () {
-    container.className = 'trap';
-  },
+  onActivate: () => container.classList.add('is-active'),
+  onDeactivate: () => container.classList.remove('is-active'),
```
Dan503, line 180: Using `classList.add` and `classList.remove` are less error prone methods of adding/removing the "is-active" class.

We should use `classList` to endorse good coding practices.

index.js
Dan503: Based on what the multiple elements with multiple traps example looks like, it doesn't look like the multi-element trap is a practical use case for animated dialogs. You are only going to animate 1 dialog at a time.

I made a test demo based on the multiple elements with multiple traps demo just to see if it would explode and it seemed to handle it ok.

Dan503:
@stefcameron the PR is ready for you to review now.

index.js
Dan503: not valid since I was basing the PR on v3 of focus trap

stefcameron:
> How is GH Pages working with the bundled js file being ignored by git though?
> Git Hub pages only publishes files if they are stored in Git. Do you have a github action that compiles it for you or something?

The repo is configured to publish pages from the root of the `gh-pages` branch. I think `gh-pages` might be an old way (maybe the original way) of doing it before GitHub added the ability to base the pages on a directory. In any case, the `bundle.js` file is not ignored on the `gh-pages` branch. If you read through the steps I pointed to for publishing the docs, you'll see what I mean.

> The GH pages change isn't really necessary. I'll revert that.

Thanks!

Dan503:
> This would mean making a PR into the gh-pages branch. Keep in mind that the code may be much older than master.

Why not just run Git Hub pages off master branch?

All you would have to do is add the bundled js to the master branch git and change your gh-pages settings to point to master. Literally 2 steps and you're done.

There are so many reasons to ditch the gh-pages branch in favour of running it on master.

1. It ensures gh-pages demo's are running on up-to-date code
2. It greatly simplifies the publishing process
3. You don't have to document the publishing process 
4. People can update code and gh-pages in the same PR
5. You don't have to switch branches to work on gh-pages so no need to stash work or anything
6. It will take less effort for you to maintain the gh-pages site

That's all I can think of right now.

stefcameron:
Relates to https://github.com/focus-trap/focus-trap-react/pull/368

stefcameron:
> > This would mean making a PR into the gh-pages branch. Keep in mind that the code may be much older than master.

I always force-push to that branch. It has no other use than to publish the demo. I don't really expect any PRs into that branch.

> Why not just run Git Hub pages off master branch?

Don't fix what isn't broken? 😄 

It's true what you say about how it would improve things. I made an issue to do this at some point. Should be pretty easy, just need the time. https://github.com/focus-trap/focus-trap/issues/408

Right now, I just barely have enough time to look at PRs across `tabbable`, `focus-trap`, and `focus-trap-react`, so the library itself is my priority.

I also really don't like this `budo` thing that runs the demos locally when trying to use them for local development of the library. I just inherited that as it's been there for ages. I really don't like that there's no way to update the library code while the demo server is running. You have to stop the server, rebuild, and restart the server to debug. It would be relatively easy to set this up with Webpack, but we already have Rollup for the builds; it would be nice if there's a way to do this with some "Rollup dev server" thing, if it exists, but again, no time and `budo` gets the job done at the moment.

index.js
```diff
@@ -470,21 +494,45 @@ const createFocusTrap = function (elements, userOptions) {
...
-        activateOptions && activateOptions.onActivate
-          ? activateOptions.onActivate
-          : config.onActivate;
-      if (onActivate) {
-        onActivate();
+      const getOption = (optionName) => {
+        return activateOptions && activateOptions[optionName] !== undefined
+          ? activateOptions[optionName]
+          : config[optionName];
+      };
```
stefcameron, line 505: This would be useful to other functions; let's promote this to the outer scope instead of keeping it in `activate()` only. No need to use it elsewhere in this PR, but `deactivate()` could make use of it, for example.

README.md
```diff
@@ -84,6 +84,8 @@ Returns a new focus trap on `element` (one or more "containers" of tabbable node
 
 - **onActivate** {function}: A function that will be called when the focus trap activates.
```
stefcameron, line 85: Let's update this doc to state that `onActivate()` is called _before_ the initial focus node gets focused.

README.md
```diff
@@ -84,6 +84,8 @@ Returns a new focus trap on `element` (one or more "containers" of tabbable node
...
 - **onDeactivate** {function}: A function that will be called when the focus trap deactivates,
+- **checkCanActivate** {function}: Animated dialogs have a small delay between when `onActivate` is called and when the focus trap is focusable. If checkCanActivate returns false, it will try sending focus again every 5 milliseconds (up to 10 seconds) until it is able to send focus.
+- **onSuccessfulActivation** {function}: This function is only relevant if used in conjunction with `checkCanActivate`. `onActivate` is called before focus is sent to the target. `onSuccessfulActivation` is called after focus is sent to the target.
```
stefcameron, line 88: Since we have `onActivate()` which is basically like a "pre-activate", how about we rename this one to `onPostActivate()` and state that it occurs after the initial focus node has received focus. It should always be called if specified, independently of `checkCanActivate()`. I don't think the two need to be tied like this.

index.js
```diff
@@ -470,21 +494,45 @@ const createFocusTrap = function (elements, userOptions) {
...
       state.active = true;
       state.paused = false;
       state.nodeFocusedBeforeActivation = doc.activeElement;
 
-      const onActivate =
-        activateOptions && activateOptions.onActivate
-          ? activateOptions.onActivate
-          : config.onActivate;
-      if (onActivate) {
-        onActivate();
```
stefcameron, line 484: I think this should remain here. After `onActivate()` has been called, then the remaining things to call are `updateTabbableNodes()` and `addListeners()` -- and now the new `onPostActivate()` -- and these are the things we want to optionally delay with `checkCanActivate()`. As it is, with your changes, we're still immediately calling `onActivate()` synchronously, we're just doing it twice, one in the IF and one in the ELSE.

index.js
```diff
@@ -470,21 +494,45 @@ const createFocusTrap = function (elements, userOptions) {
...
+          onActivate();
+        }
+
+        state.containers.forEach((container) => {
+          delayFocusTrapActivation(container, checkCanActivate, () => {
+            updateTabbableNodes();
+            addListeners();
+            if (onSuccessfulActivation) {
+              onSuccessfulActivation();
+            }
```
stefcameron, line 522: Since this is the same pattern to execute whether we delay or not, let's put this in a local function and call it in both places so we're always calling the same `finishActivation()` steps.

index.js
```diff
@@ -470,21 +494,45 @@ const createFocusTrap = function (elements, userOptions) {
...
+      const onActivate = getOption('onActivate');
+      const onSuccessfulActivation = getOption('onSuccessfulActivation');
+      const checkCanActivate = getOption('checkCanActivate');
+
+      if (checkCanActivate) {
+        if (onActivate) {
+          onActivate();
+        }
+
+        state.containers.forEach((container) => {
```
stefcameron, line 516: It's good you're thinking about the containers, but I think this will get really messy because one container may be ready and another not? What if you have 3 and two are delayed for some reason? One is ready, and the other isn't, but it'll be a race condition as it's currently implemented. You could try to manage all this, but it would get complex because of the asynchronicity.

How about we change `checkCanActivate()` to take the list of containers instead of being called once for each container?

Then the complexity of managing 2 independently delayed containers is pushed on to the consumer, who will always have better knowledge of how they want to manage that, and we don't have to deal with it here (nor do we have to test for it).

index.js
```diff
@@ -2,6 +2,30 @@ import { tabbable, isFocusable } from 'tabbable';
...
+    } else {
+      const timeDifferenceInSeconds = (Date.now() - startDate) / 1000;
+      if (timeDifferenceInSeconds > 10) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          [
+            'Focus-Trap activation for the following element timed out after',
+            timeDifferenceInSeconds,
+            'seconds',
+          ].join(' '),
```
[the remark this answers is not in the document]
stefcameron: Actually, I'm thinking a warning is not necessary here. If we were producing Dev and Prod builds, then I'd say put this type of warning in the Dev build, but we only produce Prod builds, and they typically never log to the console, so it's nice, but we should leave this out.

I'm also wondering about the interval. It feels a bit aggressive, and tailored to a use case where there's a really short delay, and they how often should you check. That's really all subjective to the consumer's scenario. For you, it's like 500ms I think, but it could be more or less for someone else.

Is there a way we could change `checkCanActivate()` such that the caller can dictate when the next check occurs, and we use `setTimeout()` instead? The method returns a boolean, but maybe it could return the delay as a number in milliseconds before the check again? If it returns anything that is "not a number or `isNaN()`", then consider it "done" and proceed with activation. If you don't specify the option, then it's `undefined`, which is the equivalent of

```javascript
const result = checkCanActivate(containers);
if (typeof result !== 'number' && !isNaN(result)) {
  finishActivation()
} else {
  setTimeout(result, delayActivation.bind(null, containers, checkCanActivate, callback));
}
```
}

What do you think?

stefcameron (changes requested):
@Dan503 Nice work on this. I do have some suggestions/requests, but this should ultimately be a nice addition to the API!

index.js
Dan503: I tried using it in `deactivate()` but it broke the return focus to origin functionality.

`deactivateOptions` has `returnFocus` but `config` has `returnFocusOnDeactivate`.

That's why I made this commit
https://github.com/focus-trap/focus-trap/pull/407/commits/d83bddee8fb73dd6fb85f8440580a7268ab504df

I can take another look at the function and see how I can make it more reusable.

stefcameron: I see, in that case, don't worry about it. This is very minor.

README.md
Dan503: > I don't think the two need to be tied like this.

In the original functionality, `updateTabbableNodes()` is called _before_ `onActivate()`.

`updateTabbableNodes()` is the thing that sends focus to the focus-trap.

If we move `onActivate()` to go before `updateTabbableNodes()` it has a high chance of breaking existing functionality for people.

I made this `if` split to retain backwards compatibility.

We could add an `onBeforeActivate()` setting but that would make it very confusing around when `onActivate()` is actually triggered. 

Maybe `onActivate()` becomes deprecated in favour of `onBeforeActivate()` and `onAfterActivate()`?

index.js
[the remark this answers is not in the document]
Dan503: Fair point about the customizable timing, I can see developers wanting to customize how frequently the checks are made.

If we remove this warning completely then how will developers know that they have an infinite loop bug and what is causing it? Suddenly performance on their site takes a massive hit and they will have no idea why.

Prod builds can have errors and warnings in the console. Website users don't see them and I'm certain that developers would appreciate knowing exactly why their site's performance has suddenly taken a nose dive if they write a bug in their code that causes an infinite loop.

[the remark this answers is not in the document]
Dan503: > For you, it's like 500ms I think, but it could be more or less for someone else.

It's actually only a few milliseconds for me. 

Yes the full animation takes 500ms but I only need to delay the activation long enough for `visibility` to transition from `hidden` to `visible` which is only a few milliseconds.

README.md
Dan503: Not quite true though unless we make a potentially breaking change by moving `onActivate()` before `updateTabbableNodes()`

index.js
Dan503: I would love to but we need to be comfortable with the idea that it will potentially break some existing code bases since it swaps the call order of `updateTabbableNodes()` and `onActivate()` around.

index.js
Dan503: I've tried to implement this but it's going to need a rather large refactor of the whole project for it to work since `updateTabbableNodes()` and `addListeners()` run across all containers all at once. You can't run `updateTabbableNodes()` based on a specific container.

This is an extremely niche use case. For initial release of the feature can we perhaps restrict the activation check to maybe just the first container in the list instead of all the containers?

This should cover 90% or more of the use cases.

You can see my failed attempt in this commit:
https://github.com/focus-trap/focus-trap/pull/407/commits/5a8a7c29eb7cd91e2aafa50cf7b0507f549d1362

demo/index.html
stefcameron: Still wondering about this. To reduce the changes in the PR that aren't directly related to adding the new feature you're working on, I'd like to revert this change from `arial-label` to `title`, even if it may be technically correct. My aim is not to provide a fully-accessible demo anyway, and as I pointed out in my previous comment on this thread, the `aria-label` already repeats the content of the button label in all cases, so there's no information being lost to someone potentially using a screen reader to try to navigate the demo page (and I don't think they would be successful today _anyway_, in spite of the `aria-label` or `title`).

demo/index.html
```diff
@@ -34,11 +34,52 @@ <h2 id="default-heading">default behavior</h2>
...
+      <p>
+        <a href="https://github.com/focus-trap/focus-trap/blob/master/demo/js/default.js">View demo source ⋙</a>
+      </p>
+    </div>
+
+    <div id="demo-animated">
+      <h2 id="animated-heading">animated</h2>
+      <p>
+        This focus trap will fade in when activated and fade out when deactivated. Note that the fade animation needs to
+        be written yourself, it is not built into the library.
```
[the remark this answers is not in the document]
stefcameron: Please add the note since it's kind of part of the demo.

demo/index.html
```diff
@@ -34,11 +34,52 @@ <h2 id="default-heading">default behavior</h2>
...
+          position: absolute;
+          opacity: 0;
+          visibility: hidden;
+          transition: opacity 0.5s, visibility 0.5s;
+        }
+        .DemoAnimatedTrap.is-active {
+          opacity: 1;
+          visibility: visible;
+        }
+      </style>
```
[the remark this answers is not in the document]
stefcameron: Please move the styles out.

demo/index.html
```diff
@@ -65,11 +106,14 @@ <h2 id="iene-heading">initial element, no escape</h2>
...
         </p>
         <p>
-          <button id="deactivate-iene" aria-label="deactivate trap for 'initial element, no escape' demo">
+          <button id="deactivate-iene" title="deactivate trap for 'initial element, no escape' demo">
             deactivate trap
           </button>
         </p>
       </div>
+      <p>
+        <a href="https://github.com/focus-trap/focus-trap/blob/master/demo/js/initial-element-no-escape.js">View demo source ⋙</a>
```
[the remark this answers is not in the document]
stefcameron: Please use some HTML character code instead of an extended character.

index.d.ts
```diff
@@ -13,6 +13,24 @@ declare module 'focus-trap' {
...
      */
     onActivate?: () => void;
+    /**
+     * A function for determining if it is safe to activate the focus trap
+     * or not. If this returns false, it will attempt to activate again in
+     * 5 milliseconds. It will repeat this up to a maximum of 10 seconds.
+     * The purpose of this is to prevent the focus-trap from activating
+     * early when dealing with animated focus-traps like dialogs that fade
+     * in and out. When a dialog fades in, there is a brief delay between
+     * the activation of the trap and the trap element being focusable.
```
stefcameron, line 23: This will likely need updating, flagging it so we don't forget.

index.js
stefcameron: I think you may have misunderstood what I was asking to change here with the multiple containers: I don't want `delayFocusTrapActivation()` to do anything with the containers. It should treat it as a unit and just pass the whole list to `checkCanActivate()` and let the consumer deal with the list as they see fit. That way, whether they're concerned about one or multiple containers, they can figure out how they can to handle this.

Now, here is what I think the code here should look like (remember from my other comment, `updateTabbableNodes()` is not the method that ultimately sets focus to the configured/defaulted "initial focus node" (see `getInitialFocusNode()`) -- it's `addListeners()` that does that).

Also note that, in order to keep the original code as unmodified as possible, I think `checkCanActivate` should be renamed `checkCanFocus`, because that's really what we're offering to delay. Otherwise, we'd have to also move the point at which `state.active = true; state.paused = false; state.nodeFocusedBeforeActivation = doc.activeElement` are called (moving that last assignment, especially, could have repercussions we don't know about), and I'd prefer to be as surgical as possible here.

```javascript
    activate(activateOptions) {
      if (state.active) {
        return this;
      }

      updateTabbableNodes();

      state.active = true;
      state.paused = false;
      state.nodeFocusedBeforeActivation = doc.activeElement;

      const onActivate = getOption(activateOptions, 'onActivate');
      const onPostActivate = getOption(activateOptions, 'onPostActivate');
      const checkCanFocus = getOption(activateOptions, 'checkCanFocus');

      if (onActivate) {
        onActivate();
      }

      const finishActivation = () => {
        addListeners();
        if (onPostActivate) {
          onPostActivate();
        }
      };

      if (checkCanFocus) {
        // would be even more succinct to use `finally(finishActivation)` but I'm leary
        //  of someone on some old version of some browser that doesn't support
        //  it since it wasn't part of the original Promises spec
        checkCanFocus(state.containers).then(finishActivation, finishActivation);
        return this;
      }

      finishActivation();
      return this;
    },
```

And now we don't even need a `delayFocusTrapActivation()` helper function!

I've also been thinking that `checkCanFocus()` should return a promise (as you can see above) and completely do away with the complexity of any sort of timer. The consumer who handles `checkCanFocus()` is then free to use whatever method they want, at whatever frequency they way, to determine the best time to activate the trap, and then resolve the promise. `focus-trap` will just wait indefinitely until the promise is resolved (or rejected) before it finishes the activation sequence and sets the initial focus.

We would need to document the fact that the `checkCanFocus` is not supported on any version of IE because of the use of a Promise (which is fine since it's new, and all but IE 11 have EOL'ed already, and this is a niche feature anyway; IE 11 EOLs on June 15, 2022; can't wait!), and also that rejecting the promise will not prevent the initial focus from taking place. The promise only serves as a way to delay setting that focus.

And now we end-up with a bonus `onPostActivate()` that someone can use to be notified _after_ the initial focus has been set and all the activation ceremonies have been done.

stefcameron (changes requested):

Dan503:
Sorry I didn't complete all your feedback before your review. I ran out of time since I spent so long trying to implement the change to `delayFocusTrapActivation()`.

Your idea about the use of a Promise sounds good.

index.js
Dan503: I can try this method. 

I don't think `updateTabbableNodes()` can be called this early, particularly on the animated example, but I can try.

We might have to move it into `finishActivation()` and document it as a possible minor low risk breaking change. 99% of projects won't break.

stefcameron: Let's just call `updateTabbableNodes()` a second time in the finish handler -- if focus was delayed:

```javascript
      const finishActivation = (focusDelayed) => {
        if (focusDelayed) {
          updateTabbableNodes();
        }
        addListeners();
        if (onPostActivate) {
          onPostActivate();
        }
      };

      ...

      checkCanFocus(state.containers).then(finishActivation, finishActivation.bind(null, true));
```

This way, we only call it the one time, and at the same place, as before if `checkCanFocus` is not provided (default); otherwise, we call it again "just to be sure", once we're allowed to set focus (and we still call it before setting focus, like we were doing before).

demo/index.html
Dan503: I've had an idea about this. 

What if I do this:
`<button aria-describedby="id-of-heading">Activate trap</button>`

Now for the default example a screen reader will read out
"button activate trap default behavior"

- It doesn't give off a tooltip
- it doesn't fail the WCAG criteria 
- it gives screen readers the extra context about what trap is being activated
- it teaches developers good accessibility practices

stefcameron: In a separate PR 🙂 

But the `aria-label` here will also be read out, so wouldn't that result in "button activate trap for 'defaults' demo default behavior"?

Or you're suggesting _removing_ the `aria-label` because the button already has "activate trap" as a label, and now we'll point to the heading as the description?

We could also point to it with `aria-labelledby` in the same way (if we remove `aria-label`).

But again, in a separate PR. We've already added all the `add/removeClass()` stuff to this PR, which also has nothing to do with what this PR is really about... Technically, that stuff, and this label stuff, should go in a separate PR about "best practices and better accessibility for the demos" or something like that. Even 2 PRs.

Dan503: The label of the button is "activate trap". We don't want aria messing around with that in any way whatsoever. 

So no to aria-labelledby since that will make screen readers announce the button as "button default behaviour" instead of "button activate trap" 

Yes to removing aria-label in favour of aria-describedby.

I think I'll open up some smaller PRs to get the demo site improvements into master then we can work more on this feature.

Dan503: I've created a separate PR now to cover the demo site improvements

https://github.com/focus-trap/focus-trap/pull/411

index.js
Dan503: I've implemented what you have suggested plus I also added onDeactivate equivalents for when the trigger is the thing that is animated (much less common use case but I can imagine it being a thing).

Since I added onDeactivate equivalents I kept the `checkCanActivate` name for symmetry with `checkCanDeactivate`. 

`checkCanDeactivate` doesn't fully stop the deactivation process though, more like it just holds off on redirecting focus back to the trigger button so `checkCanDeactivate` might not be the best name for it 🤔

Dan503: Maybe `checkCanFocusTrap` and `checkCanFocusTrigger` would be better names....

Yeah I think I'll rename them to those names.

index.js
Dan503: I managed to work around the issue by checking if `checkCanFocusTrap` exists.

If yes then it waits until that resolves to call `updateTabbableNodes()`

If not, then it calls it at the same place it did before to prevent breaking peoples code.

index.js
Dan503: I think this is resolved now.

demo/index.html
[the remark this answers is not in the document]
Dan503: This has been done in my other PR https://github.com/focus-trap/focus-trap/pull/411

When that gets merged I'll merge master back into this PR and update the new demos.

README.md
Dan503: What I meant was that `updateTabbableNodes()` was the thing causing the error.

I've handled it now in a backwards compatible way.

Dan503:
@stefcameron this is ready for you to take another look at now.

It would be good if you could approve this PR first though to reduce the size of this PR https://github.com/focus-trap/focus-trap/pull/411.

demo/index.html
stefcameron: Wonderful, thank you!

README.md
```diff
@@ -82,8 +82,12 @@ Returns a new focus trap on `element` (one or more "containers" of tabbable node
 
 #### createOptions
 
-- **onActivate** `{() => void}`: A function that will be called when the focus trap activates.
-- **onDeactivate** `{() => void}`: A function that will be called when the focus trap deactivates,
+- **onActivate** `{() => void}`: A function that will be called **before** sending focus to the target element upon activation.
+- **onPostActivate** `{() => void}`: A function that will be called **after** sending focus to the target element upon activation.
+- **checkCanFocusTrap** `{(containers: Array<HTMLElement | SVGElement | string | () => HTMLElement | SVGElement>) => Promise<unknown>}`: Animated dialogs have a small delay between when `onActivate` is called and when the focus trap is focusable. `checkCanFocusTrap` expects a promise to be returned. When that promise resolves, focus will be sent to the focus-trap. Due to the lack of Promise support, `checkCanFocusTrap` is not supported in IE11 unless you provide a Promise polyfill.
```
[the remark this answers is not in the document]
stefcameron: Also, let's remove the `<unknown>` part from `Promise<unknown>`. It's just a promise, we do nothing with any value that might be returned. And let's also mention that trap activation will proceed regardless of the Promise's resolution or rejection.

README.md
```diff
@@ -82,8 +82,12 @@ Returns a new focus trap on `element` (one or more "containers" of tabbable node
...
+- **onDeactivate** `{() => void}`: A function that will be called **before** sending focus to the trigger element upon deactivation.
+- **onPostDeactivate** `{(value: unknown) => void}`: A function that will be called **after** sending focus to the trigger element upon deactivation.
```
stefcameron, line 89: We should remove the `value: unknown` here. We don't support a value there.

README.md
```diff
@@ -82,8 +82,12 @@ Returns a new focus trap on `element` (one or more "containers" of tabbable node
 
 #### createOptions
 
-- **onActivate** `{() => void}`: A function that will be called when the focus trap activates.
-- **onDeactivate** `{() => void}`: A function that will be called when the focus trap deactivates,
+- **onActivate** `{() => void}`: A function that will be called **before** sending focus to the target element upon activation.
+- **onPostActivate** `{() => void}`: A function that will be called **after** sending focus to the target element upon activation.
+- **checkCanFocusTrap** `{(containers: Array<HTMLElement | SVGElement | string | () => HTMLElement | SVGElement>) => Promise<unknown>}`: Animated dialogs have a small delay between when `onActivate` is called and when the focus trap is focusable. `checkCanFocusTrap` expects a promise to be returned. When that promise resolves, focus will be sent to the focus-trap. Due to the lack of Promise support, `checkCanFocusTrap` is not supported in IE11 unless you provide a Promise polyfill.
+- **onDeactivate** `{() => void}`: A function that will be called **before** sending focus to the trigger element upon deactivation.
```
stefcameron, line 88: This relates to the `returnFocus` / `returnFocusOnDeactivate` option. I think we should change the description here to say, "A function that will be called **before** returning focus to the node focused prior to activation (or configured with the `setReturnFocus` option) upon deactivation."

README.md
stefcameron: Also, the description, "A function that will be called **after** returning focus..."

README.md
```diff
@@ -82,8 +82,12 @@ Returns a new focus trap on `element` (one or more "containers" of tabbable node
...
+- **onPostDeactivate** `{(value: unknown) => void}`: A function that will be called **after** sending focus to the trigger element upon deactivation.
+- **checkCanFocusTrigger** `{(trigger: HTMLElement | SVGElement) => Promise<unknown>}`: An animated trigger button will have a small delay between when `onDeactivate` is called and when the focus is sent back to the trigger. `checkCanFocusTrigger` expects a promise to be returned. When that promise resolves, focus will be sent to the trigger element. Due to the lack of Promise support, `checkCanFocusTrigger` is not supported in IE11 unless you provide a Promise polyfill.
```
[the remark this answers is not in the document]
stefcameron: Also, I'd like to name this option, `checkCanReturnFocus`. That seems more appropriate to me for what it's gating, and the other options it relates to, those being `returnFocus`, `returnFocusOnDeactivate`, and `setReturnFocus`.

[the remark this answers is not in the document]
stefcameron: Also, change `Promise<unknown>` to just `Promise`, and mention that focus will be returned to the configured node whether the promise resolves or rejects.

README.md
```diff
@@ -115,6 +119,8 @@ Returns the `trap`.
 These options are used to override the focus trap's default behavior for this particular activation.
 
 - **onActivate** `{() => void | null | false}`: Default: whatever you chose for `createOptions.onActivate`. `null` or `false` are the equivalent of a `noop`.
```
stefcameron, line 121: Not sure why we have `void | null | false` here. It can just be `void` since we ignore the return value, whatever it is, and no need for "`null` or `false` are the equivalent of a `noop`." at the end.

README.md
```diff
@@ -115,6 +119,8 @@ Returns the `trap`.
...
+- **onPostActivate** `{(value: unknown) => void | null | false}`: Default: whatever you chose for `createOptions.onPostActivate`. `null` or `false` are the equivalent of a `noop`.
```
stefcameron, line 122: Not sure why we have `void | null | false` here. It can just be `void`, and no need for "`null` or `false` are the equivalent of a `noop`." at the end.

Also, remove the `value: unknown` parameter. We do nothing with that.

README.md
```diff
@@ -115,6 +119,8 @@ Returns the `trap`.
...
+- **checkCanFocusTrap** `{(containers: Array<HTMLElement | SVGElement | string | () => HTMLElement | SVGElement>) => Promise<unknown> | null | false}`: Default: whatever you chose for `createOptions.checkCanFocusTrap`. `null` or `false` are the equivalent of a `noop`.
```
stefcameron, line 123: Return type should just be `=> Promise`, and no need for "`null` or `false` are the equivalent of a `noop`." at the end.

README.md
```diff
@@ -128,6 +134,8 @@ These options are used to override the focus trap's default behavior for this pa
 
 - **returnFocus** `{boolean}`: Default: whatever you chose for `createOptions.returnFocusOnDeactivate`.
 - **onDeactivate** `{() => void | null | false}`: Default: whatever you chose for `createOptions.onDeactivate`. `null` or `false` are the equivalent of a `noop`.
+- **onPostDeactivate** `{(value: unknown) => void | null | false}`: Default: whatever you chose for `createOptions.onPostDeactivate`. `null` or `false` are the equivalent of a `noop`.
+- **checkCanFocusTrigger** `{(trigger: HTMLElement | SVGElement) => Promise<unknown> | null | false}`: Default: whatever you chose for `createOptions.checkCanFocusTrigger`.
```
stefcameron, line 138: Same here for the `value: unknown`, `void | null | false`, and "`null` or `false` are the equivalent of a `noop`." at the end as in my previous comments.

demo/js/animated-dialog.js
```diff
@@ -0,0 +1,44 @@
...
+            if (getComputedStyle(trapContainer).visibility !== 'hidden') {
+              resolve();
+              clearInterval(interval);
+            }
+          }, 5);
+        })
+    );
+    // Return a promise that resolves when all the trap containers are able to receive focus
+    return Promise.all(results).then(
+      () => 'The value that you return here will be sent to onPostActivate'
```
stefcameron, line 24: I'd prefer not to support this return value until there's a clear use case for it.

demo/js/animated-trigger.js
```diff
@@ -0,0 +1,45 @@
...
+    trigger.classList.remove('is-triggered');
+  },
+  // There is a delay between when the class is removed
+  // and when the trigger is focusable
+  checkCanFocusTrigger: (triggerButton) => {
+    return new Promise((resolve) => {
+      const interval = setInterval(() => {
+        if (getComputedStyle(triggerButton).visibility !== 'hidden') {
+          resolve(
+            'The value that you resolve the promise with will be sent to onPostDeactivate'
```
stefcameron, line 23: I'd prefer not to support this return value.

index.d.ts
```diff
@@ -10,13 +10,52 @@ declare module 'focus-trap' {
...
+    /**
+     * A function for determining if it is safe to send focus back to the trigger element.
+     *
+     * It should return a promise that only resolves once the trigger element is focusable.
+     *
+     * The purpose of this is to prevent the focus being sent to the trigger element to early
+     * if the trigger of a focus-trap is animated to fade in and out. When a trigger element fades in,
+     * there is a brief delay between the deactivation of the trap and when the trigger element is focusable.
+     */
+    checkCanFocusTrigger?: (trigger: HTMLElement | SVGElement) => Promise<unknown>
```
stefcameron, line 57: Looks good here, just change `Promise<unknown>` to `Promise<void>`

stefcameron (changes requested):
Thanks for making those updates to the API, and adding the deactivate equivalents. We're almost there.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 6.4.1
+
+### Patch Changes
+
+- 8d11e15: Improve docs and types for most options, adding `SVGElement` as a supported type of "DOM node" since it supports the `focus()` method, same as `HTMLElement`.
```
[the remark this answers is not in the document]
Dan503: I was getting a merge conflict appearing against the readme file so I merged with master to resolved the conflict.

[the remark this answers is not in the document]
stefcameron: Looks like you merged from somewhere else because this change should only be on the `changeset-release/master` branch, not actual `master`. That other branch is created automatically by the Changeset bot.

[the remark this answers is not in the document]
Dan503: Damn, I merged `changeset-release/master` instead of `master`.

I think I saw "master" in the name and saw that it was the latest commit so I merged that instead of merging actual master branch.

I'll need to redo the merge ☹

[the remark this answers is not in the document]
stefcameron: Shoot! I _think_ you might be able to get away with just reverting the changes to the CHANGELOG and package.json because that branch would've been based on `master`, which explains why you got the README and index.d.ts changes, but also got the extra stuff.

So I would look at what's on `changeset-release/master` and just selectively remove those changes from this PR instead of trying to figure out a way to revert the merge and redo the rebase.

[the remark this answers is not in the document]
Dan503: It's ok. I'll reset my branch to before the merge, make the merge again (this time against master) and cherry pick any other changes I've made. This isn't hard to do in a GUI.

I'll need to do a force-push though to get rid of the bad merge on this branch.

README.md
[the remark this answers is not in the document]
Dan503: I can't not include the `<unknown>` part

image

README.md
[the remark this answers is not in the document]
Dan503: I feel like sending focus back to the node on rejection is probably not a good idea. If the promise rejects then it probably means it isn't safe to return the node and so it might cause crashes or something if we try to return focus on rejection as well as success then it takes control away from developers. If the developer wants a promise that resolves even if something goes wrong, they can nest a promise inside a promise.

index.js
```diff
@@ -501,25 +531,46 @@ const createFocusTrap = function (elements, userOptions) {
...
+        deactivateOptions,
+        'returnFocus',
+        'returnFocusOnDeactivate'
+      );
 
-      if (returnFocus) {
-        delay(function () {
-          tryFocus(getReturnFocusNode(state.nodeFocusedBeforeActivation));
-        });
+      const finishDeactivation = (value) => {
```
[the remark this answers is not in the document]
Dan503: The idea is that a dev can pass a value in the `checkCanReturnFocus` function and make use of it in the `onPostDeactivate` function giving developers more control like I did in the demos.

You said you didn't want to support it though so I've removed it.

index.js
Dan503: Renamed to `checkCanFocusTrap` and `checkCanReturnFocus`

CHANGELOG.md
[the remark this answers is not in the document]
Dan503: This PR is branching off of `master` branch now instead of `changeset-release/master`

README.md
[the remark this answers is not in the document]
Dan503: Changed to `Promise<void>` in the documentation and .d.ts file.

README.md
Dan503: I made the return type `Promise<void>` since `Promise` on it's own isn't allowed.

Dan503:
@stefcameron I think this is ready for you to take another look at now. 🙂

README.md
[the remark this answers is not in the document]
stefcameron: `void` is good, thanks for changing it from `unknown` to `void`.

README.md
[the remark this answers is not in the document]
stefcameron: I understand. The reason why I don't want to care what the Promise does is because I don't want to go too fast with this change, which could open a can of worms we're not thinking of right now. What should happen in this case or that case if it rejects? If we don't follow with setting focus, what state does that leave things in now, especially in the activation case? You have focus set on a node that's outside the already-activated trap, so right off the bat, focus has escaped. What happens when the user clicks the TAB key? Can of worms... 🙂 

For now, I'd like the Promise to solely be a _delay mechanism_ for what is the _inevitable_, that focus will be set (activate) or returned (deactivate), the same way it would be if the promise wasn't used.

As for the `finally`, the `then` signature is `then(resolvedHandler, rejectedHandler)`. That should be all that's necessary. What you show there (in this case, since there is only one promise and we don't care if `finishActivation()` throws or not) is functionally equivalent to `then(finishActivation, finishActivation)`.

index.js
```diff
@@ -470,21 +477,38 @@ const createFocusTrap = function (elements, userOptions) {
...
+          updateTabbableNodes();
+        }
+        addListeners();
+        if (onPostActivate) {
+          onPostActivate();
+        }
+      };
+
+      if (checkCanFocusTrap) {
+        checkCanFocusTrap(state.containers.concat()).then(finishActivation);
```
stefcameron, line 507: This should be `then(finishActivation, finishActivation)`

index.js
```diff
@@ -501,25 +525,42 @@ const createFocusTrap = function (elements, userOptions) {
...
+      };
 
-      if (returnFocus) {
-        delay(function () {
-          tryFocus(getReturnFocusNode(state.nodeFocusedBeforeActivation));
-        });
+      if (checkCanReturnFocus) {
+        checkCanReturnFocus(
+          getReturnFocusNode(state.nodeFocusedBeforeActivation)
+        ).then(finishDeactivation);
```
stefcameron, line 559: Should be `then(finishDeactivation, finishDeactivation)`

index.js
```diff
@@ -501,25 +525,42 @@ const createFocusTrap = function (elements, userOptions) {
...
+            }
+          });
+        }
+      };
 
-      if (returnFocus) {
-        delay(function () {
-          tryFocus(getReturnFocusNode(state.nodeFocusedBeforeActivation));
-        });
+      if (checkCanReturnFocus) {
```
stefcameron, line 556: I think we need to honor the `returnFocus` option here and do `if (returnFocus && checkCanReturnFocus)`

index.d.ts
```diff
@@ -10,13 +10,52 @@ declare module 'focus-trap' {
...
+    /**
+     * A function for determining if it is safe to send focus back to the trigger element.
+     *
+     * It should return a promise that only resolves once the trigger element is focusable.
+     *
+     * The purpose of this is to prevent the focus being sent to an animated trigger element too early.
+     * If a trigger element fades in upon trap deactivation, there is a brief delay between the deactivation
+     * of the trap and when the trigger element is focusable.
+     */
+    checkCanReturnFocus?: (trigger: HTMLElement | SVGElement) => Promise<void>
```
stefcameron, line 57:
```suggestion
     * The purpose of this is to prevent the focus being sent to an animated trigger element too early.
     * If a trigger element fades in upon trap deactivation, there is a brief delay between the deactivation
     * of the trap and when the trigger element is focusable.
     *
     * `trigger` will be either the node that had focus prior to the trap being activated, or the result of
     *  the `setReturnFocus` option, if configured.
     *
     * This handler is __not__ called if the `returnFocusOnDeactivate` configuration option (or `returnFocus`
     *  deactivation option) is truthy.
     */
    checkCanReturnFocus?: (trigger: HTMLElement | SVGElement) => Promise<void>
```

README.md
```diff
@@ -127,7 +133,9 @@ Returns the `trap`.
 These options are used to override the focus trap's default behavior for this particular deactivation.
 
 - **returnFocus** `{boolean}`: Default: whatever you chose for `createOptions.returnFocusOnDeactivate`.
-- **onDeactivate** `{() => void | null | false}`: Default: whatever you chose for `createOptions.onDeactivate`. `null` or `false` are the equivalent of a `noop`.
+- **onDeactivate** `{() => void}`: Default: whatever you chose for `createOptions.onDeactivate`. `null` or `false` are the equivalent of a `noop`.
+- **onPostDeactivate** `{() => void}`: Default: whatever you chose for `createOptions.onPostDeactivate`. `null` or `false` are the equivalent of a `noop`.
+- **checkCanReturnFocus** `{(trigger: HTMLElement | SVGElement) => Promise<void>}`: Default: whatever you chose for `createOptions.checkCanReturnFocus`.
```
stefcameron, line 138:
```suggestion
- **checkCanReturnFocus** `{(trigger: HTMLElement | SVGElement) => Promise<void>}`: Default: whatever you chose for `createOptions.checkCanReturnFocus`. Not called if the `returnFocus` option is truthy. `trigger` is either the originally focused node prior to activation, or the result of the `setReturnFocus` configuration option.
```

stefcameron (changes requested):
@Dan503 Solid update! I think we're like 98% there. I have made a few suggestions to clarify some things, and suggesting that `checkCanReturnFocus` is _not_ called if `returnFocus` deactivation option is falsy, which would maintain current behavior.

Dan503:
@stefcameron I've looked through your suggestions and applied what made sense. I've used some different wordings to help the sentences flow a bit nicer.

> Not called if the `returnFocus` option is truthy

I think you meant to say "falsy" here. If `returnFocus` is true then the developer wants the focus-trap to return focus.

stefcameron:
> @stefcameron I've looked through your suggestions and applied what made sense. I've used some different wordings to help the sentences flow a bit nicer.

Thank you!
 
> > Not called if the `returnFocus` option is truthy
> 
> I think you meant to say "falsy" here. If `returnFocus` is true then the developer wants the focus-trap to return focus.

Yes, that's what I meant.

stefcameron (approved):
I think we have arrived to merge time! ✅ Thank you so much for diligently working through this new feature with me. I appreciate you time and effort on this.

Dan503:
Yay! 🥳

Is there a reason you haven't merged and published yet?

Dan503:
Yay! 🥳

Is there a reason you haven't merged and published yet though?

stefcameron:
> Is there a reason you haven't merged and published yet though?

I was just pausing, and got distracted. 😄 Merging now. I'll publish shortly.

stefcameron:
@all-contributors add @Dan503 for code

Issue #113, opened 2020-04-10
optionally disable `getInitialFocusNode` on `firstTabbableNode`


defnotrobbie:
In some cases, especially in modals, it's undesirable to autofocus the first tabbable node because  the first tabbable node is a close button. That's solvable by giving the modal container `tabindex="-1"` and setting `initialFocus` to a selector for that matches the modal container. 

But what if that modal may have autofocusing children, like a text input in a registration form? And for some reason you can't assign `initialFocus` to that input? In that case, the `initialFocus` option prevents the focus trap fallback behaviour of keeping an already focused element inside the focus trap contained focused. 

a solution to this case would be adding a boolean option like `autoFocusFirstTabbable`, defaulted to `true`, that when set to `false` would prevent the fallback behavior of autofocusing the first tabbable element. Essentially, this change would make possible explicit consumer autofocus behavior instead of allowing the focus trap to search for a tabbable element and focus it.

this would be the crux of the change 
```javascript
function getInitialFocusNode() {
  var node;
  if (getNodeForOption("initialFocus") !== null) {
    node = getNodeForOption("initialFocus");
  } else if (container.contains(doc.activeElement)) {
    node = doc.activeElement;
  } else {
  // update
    node =
      (config.autofocusFirstTabbableNode ? state.firstTabbableNode : null) ||
      getNodeForOption("fallbackFocus");
  // end update
  }

  if (!node) {
    throw new Error(
      "Your focus-trap needs to have at least one focusable element"
    );
  }

  return node;
}

```

mruzekw:
I've run into a use case where this feature would be useful. I'll start working on a PR for this.

stefcameron:
> I've run into a use case where this feature would be useful.

I ran into a case myself just yesterday. Looking forward to your PR!

mruzekw:
@defnotrobbie @stefcameron Alright, that PR is up! #440

stefcameron:
@mruzekw Cool, thanks! I'll check it as soon as I have a moment.

CamilleHbp:
Shouldn't this issue be closed? :)

stefcameron:
@CamilleHbp Yes! Thank you. Fixed with #440

PR #440, opened 2021-06-24
Add initialFocusOnActivate option to control initial focus when activating trap

mruzekw:
Addresses #113

There may be cases where we don't want to focus the first tabbable element when a focus trap activates.

Examples:
- Modals/dialogs
- On mobile devices where "tabbing" doesn't make sense without a connected Bluetooth keyboard

This gives us the option to initially focus the first tabbable element or not.

- [x] Source changes maintain stated browser compatibility.
- [x] Includes updated `/docs/demo-bundle.js` if source or docs code was changed.
- N/A Unit test coverage added/updated.
- [x] E2E test coverage added/updated.
- [x] Typings added/updated.
- N/A Changes to dependencies explained.
- [x] Changeset added (run `yarn changeset` locally to add one, and follow the prompts).


stefcameron (commented):
@mruzekw Wow, thank you for such a thorough update! You covered all the Ts and dotted all the Is. 😄 

I'm wondering, though, if we really need to add yet another configuration option (there are so many now), especially since this new one basically disables the `initialFocus` option.

Adding a new option, `initialFocusOnActivate`, gives me the impression it's strictly for the first activation, but really, it's disabling all cases where we use the `initialFocus` option.

The only exception is `tryFocus(state.mostRecentlyFocusedNode || getInitialFocusNode());` in the `checkFocusIn()` handler, when focus escapes, and we use it as a fallback if `state.mostRecentlyFocusedNode` isn't set for some reason. But then, I don't think we need the fallback there if you're disabling the `initialFocus` behavior, because already, you're starting with a focus trap that doesn't contain its own focus... It's now up to the developer to set focus to some node in the trap on their own.

`state.mostRecentlyFocusedNode` should end-up getting assigned when the developer sets focus to a node inside one of the trap containers (hopefully -- that would be one special case to test for).

Why not support `false` as a valid value for the `initialFocus` option instead? If it's `=== false`, then we just ignore it.

That would mean the internal `getInitialFocusNode()` function would first check
```javascript
if (config.initialFocus === false) {
  return false;
}
...
```

And then the internal `tryFocus(node)` function would simply do nothing if `node === false` (which should also terminate the recursive stack of the function even if `node` isn't initially false, and we fall into line 210 `tryFocus(getInitialFocusNode())`).

Then we'd have one less option to manage, and I think the overall changes to `index.js` would also be somewhat less than what they are now.

What do you think about that?

mruzekw:
Thanks for your thorough feedback @stefcameron :) 

That all makes sense and sounds good to me. I'll make that change today.

Would you like me to test `state.mostRecentlyFocusedNode` gets set as well?

stefcameron:
> Thanks for your thorough feedback @stefcameron :)
> 
> That all makes sense and sounds good to me. I'll make that change today.

My pleasure, looking forward to it when you have time!

> Would you like me to test `state.mostRecentlyFocusedNode` gets set as well?

Yes, that would be good to test.

mruzekw:
@stefcameron Okay, all those changes should be up! Let me know if I overlooked anything. Thanks 🙏

stefcameron:
> @stefcameron Okay, all those changes should be up! Let me know if I overlooked anything. Thanks 🙏

👀

stefcameron (approved):
Thank you for the update! This looks great. A nice, simple addition to cover another use case.

mruzekw:
Awesome! Thanks for reviewing. It doesn't look like I have merge access. Is there someone we can ping who does? Thanks.

stefcameron:
My pleasure! I will merge it, just got distracted with work. 😄

mruzekw:
Thanks! It's no problem.

stefcameron:
This is being published in 6.6.0 right now.

PR #442, opened 2021-06-28
Add function option for escapeDeactivates option

### LINKED ISSUE #126: Add `allowEscape` or another event to handle escape without deactivation
How do you feel about adding `allowEscape` to complement `escapeDeactivates`, similar to the `clickOutsideDeactivates`/`allowClickOutside` pair?

I have a use case where I want to navigate to a different page when the user clicks outside or hits escape inside a modal. If the user cancels navigation, the modal should remain open and the focus trap should remain in effect. If `allowEscape` existed, I could achieve the behavior with the following options:

```javascript
{
    clickOutsideDeactivates: true,
    escapeDeactivates: true,
    allowClickOutside: () => {
        window.location.href = 'https://www.example.com/';
        return true;
    },
    allowEscape: () => {
        window.location.href = 'https://www.example.com/';
        return true;
    }
}
```

Another option would be to add `onClickOutside` and `onEscape` events that are invoked regardless of `clickOutsideDeactivates` and `escapeDeactivates` properties. The configuration would be:

```javascript
{
    clickOutsideDeactivates: false,
    escapeDeactivates: false,
    onClickOutside: () => window.location.href = 'https://www.example.com/',
    onEscape: () => window.location.href = 'https://www.example.com/'
}
```

In any case, I'm happy to send a pull request with the implementation if I can get an agreement on what the API should look like.

**Edit:** Added another example.

stefcameron:
Fixes #126

The `escapeDeactivates` option, like other similar boolean options,
can now be a function that takes the related keyboard event and
returns a boolean, instead of just a straight boolean.

<details>
<br/>

- Includes updated `/docs/demo-bundle.js` if source or docs code was changed.
- Typings added/updated.

</details>


stefcameron:
This is being published in 6.6.0 right now.

PR #497, opened 2021-09-05
✨ setReturnFocus function improvements

### LINKED ISSUE #485: ✨ Proposal - Add `setReturnFocus` option to `deactiave` method
### What and why:
In the time of creation `focus-trap` sometimes we don't know beforehand what element should be focused after deactivation. Imagine a situation with modal and focus-trap inside. When I click outside of modal (on focusable element) I would like to deactivate trap (modal is closed) and focus on clicked element.

### How:
We don't have access to config after creation of trap, but we have `deactiaveOptions` which are passed to `deactivate` method. Inside it, we can extend current behavior. In `finishDeactivation` method, focus element should be (order makes difference):
1. from `setReturnFocus` property passed to `deactivate` method
2. from `setReturnFocus` property passed to config
3. last focused node before activation

Willing to see any alternatives.
If the above plan makes sense, I can try to bake PR.

far-fetched:
Fixes #485

Received last focused element as argument and accept false as return value

<details>
<br/>

- Typings added/updated.

</details>


docs/index.html
```diff
@@ -631,6 +631,38 @@ <h2 id="setreturnfocus-heading">setReturnFocus option</h2>
...
+        </button>
+        <p id="deactivate-setreturnfocus-function">
+          <button id="focus-this">
+            deactivate and focus this
+          </button>
+          <button id="focus-initial">
+            deactivate and focus 'activate trap'
+          </button>
+          <button id="no-focus">
+            deactivate and no focus
```
stefcameron, line 651: Could we label it "deactivate and no change in focus", since it'll just leave whatever's focused at the time of deactivation focused? Having just "no focus" makes me think it'll remove focus too, which it won't.

docs/js/set-return-focus-function.js
```diff
@@ -0,0 +1,41 @@
...
+    return previousActiveElement;
+  }
+  return false;
+};
+
+const focusTrap = createFocusTrap('#setreturnfocus-function', {
+  onActivate: () => container.classList.add('is-active'),
+  onDeactivate: () => container.classList.remove('is-active'),
+  setReturnFocus,
+  allowOutsideClick: true,
```
stefcameron, line 19: It would be nice to use the function here to limit acceptable clicks to those that are on the deactivate buttons. Otherwise, I can go activate another trap, and that may seem odd to some:

image

docs/index.html
```diff
@@ -631,6 +631,38 @@ <h2 id="setreturnfocus-heading">setReturnFocus option</h2>
       </p>
     </div>
 
+    <div id="demo-setreturnfocus-function">
+      <h2 id="setreturnfocus-function-heading">setReturnFocus as function</h2>
+      <p>
+        As the function is executed in the process of deactivation, it allows you to dynamically specify which element should be focused after deactivation.
+      </p>
```
stefcameron, line 638: Adding a second paragraph here, like "Once the trap is activated, you can still click on one of the 3 deactivate buttons below in order to deactivate the trap." would be helpful.

stefcameron (changes requested):
@far-fetched Thanks for the PR! I really appreciate it. 😄  Looking good. Just a few minor suggestions.

README.md
```diff
@@ -98,7 +98,7 @@ Returns a new focus trap on `element` (one or more "containers" of tabbable node
...
+- **setReturnFocus** `{HTMLElement | SVGElement | string | (previousActiveElement: HTMLElement | SVGElement) => HTMLElement | SVGElement | false}`: By default, on **deactivation**, if `returnFocusOnDeactivate=true` (or if `returnFocus=true` in the [deactivation options](#trapdeactivatedeactivateoptions)), focus will be returned to the element that was focused just before activation. With this option, you can specify another element to programmatically receive focus after deactivation. It can be a DOM node, a selector string (which will be passed to `document.querySelector()` to find the DOM node **upon deactivation**), or a function that returns a DOM node to call **upon deactivation** (i.e. the selector and function options are only executed at the time the trap is deactivated).
```
stefcameron, line 101: I think we need another tweak here I didn't notice the first time around, because we're also adding the option to return `false`:

```suggestion
- **setReturnFocus** `{HTMLElement | SVGElement | string | (previousActiveElement: HTMLElement | SVGElement) => HTMLElement | SVGElement | false}`: By default, on **deactivation**, if `returnFocusOnDeactivate=true` (or if `returnFocus=true` in the [deactivation options](#trapdeactivatedeactivateoptions)), focus will be returned to the element that was focused just before activation. With this option, you can specify another element to programmatically receive focus after deactivation. It can be a DOM node, a selector string (which will be passed to `document.querySelector()` to find the DOM node **upon deactivation**), a function that returns a DOM node to call **upon deactivation** (i.e. the selector and function options are only executed at the time the trap is deactivated), or `false` to leave focus where it is at the time of deactivation.
```
```

File: index.d.ts [hunk 5]
```diff
@@ -106,7 +106,7 @@ declare module 'focus-trap' {
      * By default, focus trap on deactivation will return to the element
      * that was focused before activation.
      */
-    setReturnFocus?: FocusTarget;
+    setReturnFocus?: FocusTargetOrFalse;
```
stefcameron, line 109: Now that we're passing a parameter to the option if it's a function, I think we need to adjust the typings here, otherwise, someone using TypeScript will have a hard time using the parameter.

Now I'm not a TypeScript expert, so while there may be a fancy way to express this using some type of template parameter or type tooling function, it's probably easiest to just not use `FocusTargetOrFalse` and write it all out:

```TypeScript
setReturnFocus?: HTMLElement | SVGElement | string | false | ((nodeFocusedBeforeActivation: HTMLElement | SVGElement) => HTMLElement | SVGElement | false);
```

far-fetched:
Changes addressed ;)

stefcameron (approved):
Thanks for the additional fixes, this looks great! 🎉

stefcameron:
@all-contributors add @far-fetched for docs, code, tests

PR #504, opened 2021-09-08
Add document option

wandroll:
If using in an iFrame context, we may want to specify the target document for the focus trap to work correctly.

Original author of this : https://github.com/MD4

<details>
<br/>

- Typings added/updated.

</details>


stefcameron:
So we're updating #98 to fix #97, great! Thanks for doing this.

stefcameron (changes requested):
This looks good, but we're missing two things:

1. We already an `iframe` test -- could you extend that test, or add another case, to cover this new feature?
2. Please generate a new demo-bundle.js file (run `yarn demo-bundle`) and include it in this PR.

wandroll:
> This looks good, but we're missing two things:
> 
> 1. We already an `iframe` test -- could you extend that test, or add another case, to cover this new feature?
> 2. Please generate a new demo-bundle.js file (run `yarn demo-bundle`) and include it in this PR.

Ok, I'll do that! In parrallel, I 've worked on an example in the focus-trap-react https://github.com/focus-trap/focus-trap-react/pull/460

stefcameron:
> Ok, I'll do that! In parallel, I've worked on an example in the focus-trap-react [focus-trap/focus-trap-react#460](https://github.com/focus-trap/focus-trap-react/pull/460)

Nice, thank you! I appreciate that, since we'll need at least the prop-types update over there too.

stefcameron (commented):
Thank you for adding the test and re-generating the demo bundle! Looks like there's a security error in the FF tests. Maybe there's a similar flag for FF like the `chromeWebSecurity` one you added?

[Reading-up on this](https://docs.cypress.io/guides/guides/web-security#Disabling-Web-Security), unfortunately, it's a Chrome-only thing. So no wonder FF fails.

To keep the test, I think we would have to refactor things so that we have a second `.spec.js` file with this one test in it that we only run in Chrome, which seems a bit complicated.

I'm also wondering what's up with the Chrome tests, why they have once again been canceled before they could run.

I realize I asked you to write the test -- and I do appreciate it -- but let's try disabling the test and seeing if things go through in CI. Regardless, I'm glad we have the additional demo, even if we don't have an automated test against it.

Edit: I didn't realize the iframe would cause all these security context issues...

I just don't want you to end-up spending hours and hours bashing your head against a Cypress wall. I've been there before...

stefcameron:
Thanks for disabling the test. Still the same issues, unfortunately, and seems to be in the CI env only, since I can't get Chrome to fail locally. But I can get FF to fail locally.

The problem with FF is the code in `in-frame.js` for the demo.

Maybe it's the same for Chrome, why it disconnects before the tests complete. Maybe it's the async checking for the loading of the iframe content.

In any case, if I comment the code in `in-frame.js`, then FF succeeds.

stefcameron:
Locally, everything works if I also just comment out `require('./in-iframe');` in `docs/js/index.js`. So I'm thinking if there was a way to conditionally require that only if cypress is not running the demo, then we could keep the demo for manual demoing/testing. Just not sure how to do that, or if it's possible. I think it would ultimately be a Browserify thing, but I've never used Browserify before (just inherited it when I took over this repo).

stefcameron:
I think I may have a solution, using https://www.npmjs.com/package/envify

stefcameron:
I think I have a solution, but for some reason, I'm unable to push to your fork. Normally, I can push to forks of a repo I maintain. 🤔

stefcameron:
I had to fork your fork and make this PR into it: https://github.com/iadvize/focus-trap/pull/1

When you have a moment, could you merge that and then we can see if I solved the issue in GitHub CI?

wandroll:
> I think I have a solution, but for some reason, I'm unable to push to your fork. Normally, I can push to forks of a repo I maintain. 🤔

I guess its because its a company repo.

stefcameron:
Success! 🎉

stefcameron (approved):
Good to go!

stefcameron:
@all-contributors add @wandroll for code, test, docs, demo

stefcameron:
@wandroll Thank you for doing this!

PR #516, opened 2021-09-17
[#496] Support focus traps inside open shadow DOMs

### LINKED ISSUE #496: Problem inside shadow dom
Hello!

Consider a `<div>` to which a shadow dom is attached. 

If I then add elements below this and use focus-trap for a child element (modal dialog im my case) the click event on the shadow dom will trigger for all those child elements and the event will be cancelled by this code:

```
  var checkClick = function checkClick(e) {
    if (valueOrHandler(config.clickOutsideDeactivates, e)) {
      return;
    }

    if (containersContain(e.target)) {
      return;
    }

    if (valueOrHandler(config.allowOutsideClick, e)) {
      return;
    }

    e.preventDefault();
    e.stopImmediatePropagation();
  }; //

```

If I change
`    if (containersContain(e.target)) {
`
to
`    if (containersContain(e.path[0])) {
`
it works.
path[0] will always be the actual clicked event, while the target in this case is the shadow dom parent.

stefcameron:
Fixes #496

trap-in-shadow

<details>
<br/>

- Typings added/updated.

</details>


stefcameron:
@simonihmig 
> Besides the changes here, wouldn't you also need to change the default assignment of `doc`? As you cannot access elements behind the shadow root when using `document`. I would expect these things to fail as they will just return the shadow host (similar to `event.target`) and not the actual element behind the shadow DOM IIRC:

That might explain my failing tests... 😆 Thanks for bringing this up!

> I saw you can now pass a `document` option, so the user could pass his shadow root here. But it would be more convenient if this would work out of the box. Which _should_ be possible I think, as you can get either the shadow root (if present) or just the regular `document` by calling [`element.getRootNode()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode).

Yes, we just added (not published yet) the `document` option, but that was mainly for iframes. I wonder if `element.getRootNode()` would work for that as well... Doesn't look like it, based on the docs, but looks like it would be quite useful for shadow DOM stuff like this.

I'll check this out, thank you!

stefcameron:
OK, I've got this working now, along with tests:

trap-in-shadow

@simonihmig I had a look at your suggestion of using `element.getRootNode()` when setting the internal `doc` variable used by focus-trap (which you can now configure using the `document` option).

You can do this, and then the `doc.querySelector()` and `doc.activeElement` calls will work, however, focus won't be prevented outside the trap (i.e. outside that shadow DOM). So while it'll be trapped within the container within your shadow DOM, you'll be able to click on anything in the host, because focus-trap will add all its doc-based listeners to the shadow host and listen inside of it, rather than listening for clicks and such on the host.

The only drawback to __not__ using `element.getRootNode()` is that the 2 options that accept selector strings to identify nodes won't work if you give a selector to an element inside your shadow DOM for reasons you stated earlier. So as long as you can avoid using selector strings (or don't need to use those 2 options), then you should be fine.

I documented this in the README as well as in the demo page (in this PR).

stefcameron:
@andersthorsen was able to validate this works in his env where he originally experienced the bug, so merged!

simonihmig:
@stefcameron I can confirm as well, works fine now for me! :tada:  
Thank you!

stefcameron:
> @stefcameron I can confirm as well, works fine now for me! 🎉
> Thank you!

🚀 Awesome, thank you!

PR #588, opened 2022-01-08
[#565] Fix issue with last element with negative tabindex in trap

### LINKED ISSUE #565: Tab has no effect if tabindex=-1 element after other tabbable elements is focused
Hi,

If there is an element with tabindex=-1, it is focused, and it is the last element within the focus trap, pressing the tab key has no effect.

I've created a small demo here (with focus-trap 6.7.1):

https://codesandbox.io/s/focus-trap-tab-index-forked-gynbe

To reproduce:

1. Activate the trap
2. Use the mouse to focus the last button
3. Press the tab key

The focus doesn't move anywhere, while I would expect it to move back to the first button in the trap.

Shift+Tab isn't affected and tabs back as expected. Also, if the tabindex=-1 element isn't last, everything works as expected.

(Note: I'm finding codesandbox.io a bit temperamental – I've also put it on Netlify here as an alternative: https://csb-gynbe.netlify.app/)

# Demo

focus-trap

stefcameron:
Fixes #565

negative-tabindex

<details>
<br/>

  - ⚠️ Non-covered demos (look for `IS_CYPRESS_ENV === ''` [here](https://github.com/focus-trap/focus-trap/blob/master/docs/js/index.js), as well as `in-open-shadow-dom.js` that can't be fully tested in Cypress) __manually__ verified.
- Typings added/updated.
- Changes do not break SSR:

</details>


stefcameron:
The "multiple traps with multiple elements works" test is failing because the following statement is doing __nothing__:

```javascript
      cy.get('@testRoot')
        .findByRole('button', { name: /^deactivate trap 2/ })
        .click();
```

The button is found, the click is made, but the click handler is never called (to be sure, I put a `throw new Error("foo")` statement in the button's click handler, and the error isn't thrown), hence the trap doesn't deactivate, and the following test fails.

I see zero reason for this, other than apparent Cypress coin flipping. So frustrating. The demo works fine manually.

jpveooys:
@stefcameron I'm not sure why but adding using `.click({force: true})` for the two deactivate clicks in that test seems to fix the problem. (It's odd as usually Cypress complains if the click fails.)

stefcameron:
> @stefcameron I'm not sure why but adding using `.click({force: true})` for the two deactivate clicks in that test seems to fix the problem. (It's odd as usually Cypress complains if the click fails.)

Interesting. Thank you for the suggestion! Trying now. Cypress is not my strong suit.

Issue #375, opened 2021-04-16
focus-trap stuck on elements with positive tabIndex


amje:
Hi!

It seems elements with positive tabIndex break tab/shift+tab navigation. When such elements receive focus moving forward become unavailable. While moving backward returning to that element is also unavailable, focus cycle inside container is not working.

Here's a repro example - https://codesandbox.io/s/focus-trap-tab-index-u1fxu

**Current behavior**
1. Activate the trap -> focus moves to button with tabIndex="1"
2. Press Tab -> nothing happens

**Expected behavior**
1. Activate the trap -> focus moves to button with tabIndex="1"
2. Press Tab -> focus moves to button with no tabIndex

stefcameron:
@amje Thanks for reporting this, and for the sandbox repro. I will check this out!

stefcameron:
Indeed, this is strange behavior.

stefcameron:
I did some investigating and identified the problem in branch `issue375`. To be continued...

gotbahn:
Can reproduce this as well

stefcameron:
Thanks for the input. I've got something in the works, but I think it's going to be quite a large change so may take a while for me to get through it.

stefcameron:
Slow going, but I just had an idea of how to solve this, so I updated my branch and included my idea in a `// DEBUG IDEA` comment: https://github.com/focus-trap/focus-trap/blob/issue375/index.js#L298

Will get back to this again later. Unless someone wants to run with this idea and make a PR out of my branch. Just LMK if you're working on it.

DaviDevMod:
@stefcameron I am writing a focus trap and I solved this problem. I can't link to my code because is not on GH yet (the repo is a mess), but I'll leave a snippet in here. Anyway the algorithm is what matter the most. 
Unfortunately it's quite a bit of logic to add, just to support those unlikely non-zero tab indexes, but here we are.

One needs to find the firstZero, lastZero, topTabbable and bottomTabbable in each container.
Where top/bottom are simply the first and last tabbable in document order, while first and last zero are the first and last zero tab indexes in document order.

And also create an array with all the positive tab indexes in the trap, ordered by tab index, resolving ties by document order.

(I know that focus-trap outsources the search for tabbable elements to focus-trap/tabbable, so you don't have that much freedom of movement here, but still can get the desired results, manipulating the output of focus-trap/tabbable)

So then when a tab event occurs if:
[...]

PR #616, opened 2022-02-06
Fix bug with negative tabindex NOT last in container

### LINKED ISSUE #611: Focusing an isFocusable, yet not isTabbable element results in focusing the first element in the container
Hi there!

We are using skip links in our modal which leverages the focus-trap. It seems that the way #565 was fixed broke our skip link implementation :)

It basically looks like this:
```
<hr id="skip-link-target-top"  />         # visibility: hidden
<a href="#skip-link-target-bottom" />
...
<hr id="skip-link-target-bottom" />       # visibility: hidden
<a href="#skip-link-target-top" />
```

When the user clicks (tabs I should say - it's actually hidden as well) on the skip link, the corresponding target becomes visible and gets focused.

I created a codesandbox to showcase the behaviour: https://codesandbox.io/s/focus-trap-skip-links-eyz0g
It works with version 6.7.1 but not with version 6.7.2.

Basically https://github.com/focus-trap/focus-trap/blob/1700055ed7629aecbd5b8fb73cd93dc0f248cffc/index.js#L414 leads to `lastOfGroupIndex` to be set to `containerIndex` which a few lines later on sets the `destinationNode` to `destinationGroup.firstTabbableNode`.

I was wondering if we should try a different approach to improve our UX for both sighted and blind users, but on the otherhand, the underlying assumption which is made on line 418ff doesn't seem right to me? :thinking: From a userperspective I would expect the tabflow to continue from the focused element.

What do you think?

Best regards!

stefcameron:
Fixes #611

negative-tabindex-middle

<details>
<br/>

  - ⚠️ Non-covered demos (look for `IS_CYPRESS_ENV === ''` [here](https://github.com/focus-trap/focus-trap/blob/master/docs/js/index.js), as well as `in-open-shadow-dom.js` and `negative-tabindex.js` that can't be fully tested in Cypress) __manually__ verified.
- Typings added/updated.
- Changes do not break SSR:

</details>


stefcameron:
There's a lint error in here on purpose because this needs further testing. It also needs a Changeset.

stefcameron:
@skriems Please have a look... 🙂

skriems:
Hey @stefcameron, that looks pretty good to me!

I haven't tested the solution really but since all of this is done inside `updateTabbableNodes` it should work regardless if the element got `tabindex=-1` via javascript or not. _brilliant_ :tada: :smile: 

Regarding the positive tabindex values... that's much much harder of a problem.... :thinking: I'll definately think about it!

Thanks for looking into that and your time!

stefcameron:
> Hey @stefcameron, that looks pretty good to me!
> 
> I haven't tested the solution really but since all of this is done inside `updateTabbableNodes` it should work regardless if the element got `tabindex=-1` via javascript or not. _brilliant_ 🎉 😄

Glad to know it's looking good to you! You definitely inspired the fix.
 
> Regarding the positive tabindex values... that's much much harder of a problem.... 🤔 I'll definately think about it!

I'm concerned that the fix here won't work in the case of positive tabindex. We already have an outstanding issue with those on #375. But I guess whether this fix goes in or not, positive tabindex would still be broken because moving the focus to the first tabbable node in the container would be just as wrong as miscalculating the tab order in the negative tab index case.

> Thanks for looking into that and your time!

My pleasure.

stefcameron:
I've come to the conclusion that since positive tabindex is already broken (see #375), this fix isn't going to make any matters worse, so we might as well go with it.

Issue #682, opened 2022-05-02
Setting initialFocus=true leads to a call stack exceeded exception


stefcameron:
Because we're not handling `true` as a possible value. Docs state that only `false` is supported, but code still needs to guard against `true` being given.

<img width="490" alt="image" src="https://user-images.githubusercontent.com/2855350/166310617-6fb06721-1a23-43d8-880c-e6ab195da694.png">

far-fetched:
Hey @stefcameron
I can take care of this issue, if you don't mind.
How `trap` should behave when true is passed :thinking: Do we need any special handling or just throw an exception that this value is not supported ?

stefcameron:
@far-fetched Thank you, I greatly appreciate that! I think we should interpret `true` as signaling that the user wants the default value, which is, "By default, when a focus trap is activated the first element in the focus trap's tab order will receive focus." That basically means ignoring `true` and doing what we would normally do if the option wasn't set at all.

PR #691, opened 2022-05-09
Handle unexpected params passed

far-fetched:
fixes: #682 
PR adds handle for unexpected value – `true` (not mentioned in docs).

Initial issue was concern only for `initialValue`. During tests, I figured out that `setReturnFocus` affects similar behavior. 
When passing `setReturnFocus=true` it does not return to previously focused node (default behavior).
Because of above, I decided to add handle fix in [getNodeForOption](https://github.com/far-fetched/focus-trap/blob/7a966c94db747a01918c7d12fd5d270e4a8d8ddd/index.js#L198) function. It is used to get `initialValue`, `setReturnFocus` and `fallbackFocus`. 
So this change fixes problems with passing true for:

- `initialValue` – mentioned in related issue

- `setReturnFocus` – use default behavior if true

- `fallbackFocus` – I haven't noticed any change and tests are green

Hey @stefcameron would like to hear your feedback, thanks in advance
I am not sure about adding tests. I can add them to prevent future regression, but as we used them as demo, for some users it might suggest that you have to pass e.g., `initialValue=true` – which is not mandatory. 

PS: I understand demo-bundle is generated automatically, but it seems that it adds some strange changes, should I commit them ?

<details>
<br/>

  - ⚠️ Non-covered demos (look for `IS_CYPRESS_ENV === ''` [here](https://github.com/focus-trap/focus-trap/blob/master/docs/js/index.js), as well as `in-open-shadow-dom.js` and `negative-tabindex.js` that can't be fully tested in Cypress) __manually__ verified.
- Typings added/updated.
- Changes do not break SSR:

</details>


.changeset/modern-peas-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'focus-trap': patch
+---
+
+handle unexpected param (true) passed as initialValue option
```
stefcameron, line 5:
```suggestion
Handle unexpected param (true) passed as the value for the `initialFocus`, `fallbackFocus`, and `setReturnFocus` options: Ignore and perform default behavior.
```

stefcameron (changes requested):
Thanks for fixing this issue -- and glad you saw that it was affecting two other options! 🎉 

> Hey @stefcameron would like to hear your feedback, thanks in advance
I am not sure about adding tests. I can add them to prevent future regression, but as we used them as demo, for some users it might suggest that you have to pass e.g., initialValue=true – which is not mandatory.

Yeah, I think in this case, we can forego adding tests. I don't want to mislead users in thinking that `true` is supported, and if these options someday start supporting `true` to mean something specific, we'll find out soon enough that this function is overwriting that case and doing the default behavior.

stefcameron (approved):
Thank you for this fix!

PR #816, opened 2022-10-26
expose activeFocusTraps

vinicius73:
Expose `activeFocusTraps` to make able to deal with multiple versions of `focus-trap` in the same page.

```js
window.__trapQueue = window.__trapQueue ?? []

const activeFocusTraps = (function () {
  const trapQueue = window.__trapQueue;

  return {
    activateTrap(trap) {
      // --
    },

    deactivateTrap(trap) {
      // --
    },
  };
})();

const focusTrap = createFocusTrap(container, {
  activeFocusTraps,
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ Returns a new focus trap on `element` (one or more "containers" of tabbable node
 - **preventScroll** `{boolean}`: By default, focus() will scroll to the element if not in viewport. It can produce unintended effects like scrolling back to the top of a modal. If set to `true`, no scroll will happen.
 - **delayInitialFocus** `{boolean}`: Default: `true`. Delays the autofocus to the next execution frame when the focus trap is activated. This prevents elements within the focusable element from capturing the event that triggered the focus trap activation.
 - **document** {Document}: Default: `window.document`. Document where the focus trap will be active. This allows to use FocusTrap in an iFrame context.
-- **activeFocusTraps**: (optional) `{ activateTrap(trap: FocusTrap) => void, deactivateTrap(trap: FocusTrap) => void }` Enables to manualy handle multiple focus-trap instances.
+- **activeFocusTraps**: (optional) `{ activateTrap(trap: FocusTrap) => void, deactivateTrap(trap: FocusTrap) => void }` Enables to manually handle multiple focus-trap instances.
 - **tabbableOptions**: (optional) [tabbable options](https://github.com/focus-trap/tabbable#common-options) configurable on FocusTrap (all the *common options*).
   - ⚠️ See notes about **[testing in JSDom](#testing-in-jsdom)** (e.g. using Jest).
 
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ Returns a new focus trap on `element` (one or more "containers" of tabbable node
 - **preventScroll** `{boolean}`: By default, focus() will scroll to the element if not in viewport. It can produce unintended effects like scrolling back to the top of a modal. If set to `true`, no scroll will happen.
 - **delayInitialFocus** `{boolean}`: Default: `true`. Delays the autofocus to the next execution frame when the focus trap is activated. This prevents elements within the focusable element from capturing the event that triggered the focus trap activation.
 - **document** {Document}: Default: `window.document`. Document where the focus trap will be active. This allows to use FocusTrap in an iFrame context.
+- **activeFocusTraps**: (optional) `{ activateTrap(trap: FocusTrap) => void, deactivateTrap(trap: FocusTrap) => void }` Enables to manualy handle multiple focus-trap instances.
 - **tabbableOptions**: (optional) [tabbable options](https://github.com/focus-trap/tabbable#common-options) configurable on FocusTrap (all the *common options*).
   - ⚠️ See notes about **[testing in JSDom](#testing-in-jsdom)** (e.g. using Jest).
 
--- a/docs/js/custom-active-focus-traps.js
+++ b/docs/js/custom-active-focus-traps.js
@@ -1,0 +1,25 @@
+const { createFocusTrap } = require('../../index');
+
+module.exports = () => {
+  const container = document.getElementById('custom-active-focus-traps');
+
+  const activeFocusTraps = {
+    activateTrap: (trap) => {
+      container.classList.add('is-active');
+    },
+    deactivateTrap: (trap) => {
+      container.classList.remove('is-active');
+    },
+  };
+
+  const focusTrap = createFocusTrap(container, {
+    activeFocusTraps,
+  });
+
+  document
+    .getElementById('custom-active-focus-traps-activate')
+    .addEventListener('click', focusTrap.activate);
+  document
+    .getElementById('custom-active-focus-traps-deactivate')
+    .addEventListener('click', focusTrap.deactivate);
+};
--- a/cypress/e2e/focus-trap-demo.cy.js
+++ b/cypress/e2e/focus-trap-demo.cy.js
@@ -1531,6 +1531,58 @@ describe('focus-trap', () => {
     });
   });
 
+  describe('demo: custom focus-traps', () => {
+    it('traps focus tab sequence and allows deactivation by clicking deactivate button', () => {
+      cy.get('#demo-with-custom-active-focus-traps').as('testRoot');
+
+      // activate trap
+      cy.get('@testRoot')
+        .findByRole('button', { name: /^activate trap/ })
+        .as('lastlyFocusedElementBeforeTrapIsActivated')
+        .click();
+
+      // 1st element should be focused
+      cy.get('@testRoot')
+        .findByRole('link', { name: 'with' })
+        .as('firstElementInTrap')
+        .should('be.focused');
+
+      // trap is active(keep focus in trap by blocking clicks on outside focusable element)
+      cy.get('#return-to-repo').click();
+      cy.get('@firstElementInTrap').should('be.focused');
+
+      // trap is active(keep focus in trap by blocking clicks on outside un-focusable element)
+      cy.get('#demo-with-custom-active-focus-traps-heading').click();
+      cy.get('@firstElementInTrap').should('be.focused');
+
+      // trap is active(keep focus in trap by tabbing through the focus trap's tabbable elements)
+      cy.get('@firstElementInTrap')
+        .tab()
+        .should('have.text', 'some')
+        .should('be.focused')
+        .tab()
+        .should('have.text', 'focusable')
+        .should('be.focused')
+        .tab()
+        .as('lastElementInTrap')
+        .should('contain', 'deactivate trap')
+        .should('be.focused')
+        .tab();
+
+      // trap is active(keep focus in trap by shift-tabbing through the focus trap's tabbable elements)
+      cy.get('@firstElementInTrap').should('be.focused').tab({ shift: true });
+      cy.get('@lastElementInTrap').should('be.focused');
+
+      // focus can be transitioned freely when trap is deactivated
+      cy.get('@testRoot')
+        .findByRole('button', { name: /^deactivate trap/ })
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -118,6 +118,7 @@ Returns a new focus trap on `element` (one or more "containers" of tabbable node
 - **preventScroll** `{boolean}`: By default, focus() will scroll to the element if not in viewport. It can produce unintended effects like scrolling back to the top of a modal. If set to `true`, no scroll will happen.
 - **delayInitialFocus** `{boolean}`: Default: `true`. Delays the autofocus to the next execution frame when the focus trap is activated. This prevents elements within the focusable element from capturing the event that triggered the focus trap activation.
 - **document** {Document}: Default: `window.document`. Document where the focus trap will be active. This allows to use FocusTrap in an iFrame context.
+- **activeFocusTraps**: (optional) `{ activateTrap(trap: FocusTrap) => void, deactivateTrap(trap: FocusTrap) => void }` Enables to manualy handle multiple focus-trap instances.
```
nickvergessen, line 121:
```suggestion
- **activeFocusTraps**: (optional) `{ activateTrap(trap: FocusTrap) => void, deactivateTrap(trap: FocusTrap) => void }` Enables to manually handle multiple focus-trap instances.
```

stefcameron:
@vinicius73 Could you please explain your use case? I don't understand why something like this is necessary. I'm also not really liking putting stuff in the global object.

vinicius73:
> @vinicius73 Could you please explain your use case? I don't understand why something like this is necessary. I'm also not really liking putting stuff in the global object.

In `nextcloud` we have [`@nextcloud/vue`](https://github.com/nextcloud/nextcloud-vue) who implements `focus-trap`.
A lot of officials and community apps use this package.

All apps are shipped with all standalone libraries, so if We have 2 apps installed, and those apps are using `@nextcloud/vue`, We will have 2 "isolated instances" of `focus-trap`, for example.

Because of that, every focus-trap "instance" have `activeFocusTraps` "singleton state", who not interact with each other.

In a "common" application, it is not a problem, but in our case, we need "share" the active instances with a unique and global state.

---

We have already tried it. https://github.com/nextcloud/nextcloud-focus-trap/blob/72e1fda5ba4a7a1f2a519ebab446cf4db7045d8e/lib/index.ts#L7-L15, unfortunately it was not sufficient, since there is an internal state in `focus-trap ` who we can't interact.

vinicius73:
> I'm also not really liking putting stuff in the global object.

The comment is an example of our need @stefcameron, it is not in source code at all.

The implementation is the only way to have a shared state across all installations of focus-trap in or platform.

skjnldsv:
@stefcameron if you want to have a chat or more in-depth discussion about this, we can also invite you to a discussion channel :)
Thanks for the help! :hugs:

stefcameron:
@vinicius73 Thank you for the explanation of your use case, and @skjnldsv thank you for the invitation to your chat.

While I appreciate your proactiveness in posting a PR with a possible solution, I'm not in favor of taking this in because focus-trap has good reason to have its internal state, and because I feel the real problem here is that NextCloud is built in such a way that you end-up with multiple, possibly differing versions of focus-trap competing for focus in the browser.

I had a look at NextCloud and found that `@nextcloud/focus-trap` is likely the issue because it declares a hard dependency on `focus-trap`: https://github.com/nextcloud/nextcloud-focus-trap/blob/master/package.json#L40-L42

Instead, as a library, it should declare a __peer__ dependency, and leave it to the app author to add focus-trap to their project so that the app they build never ends up with more than one instance of focus-trap used both by the app, and by the component library.

Otherwise, I feel that your proposed solution here is more of a hack around this packaging/bundling issue, and will place undue burden/expectations on this library for maintenance going forward. Especially in the case where differing versions of focus-trap are used in the same state, one version might have features the other one doesn't, or might have a bug that's been fixed in the other. Combining states between different versions (even the same version, honestly) seems like a recipe for strange behavior that I don't want to attempt to support.

Can you request that `@nextcloud/focus-trap` mark focus-trap as a peer instead? Or have I completely misunderstood the problem?

vinicius73:
Thanks for the answer @stefcameron, but unfortunately it will not solve the problem.
As I've mentioned, every app is bundled separated without share any code at browser.

It is not about "multiple versions" of focus trap, even if every app uses the same version of focus-trap, we will have multiple "separated and isolated" instances of focus-trap code in every app.

skjnldsv:
@stefcameron the issue is not with Nextcloud.
If two apps built in different environments both ships their usage of `focus-trap`, they will face the same problem.

You can ignore `@nextcloud/focus-trap`, this is a try by us to mitigate the issue.
We initially just used `focus-trap` on our apps and components.

stefcameron:
> It is not about "multiple versions" of focus trap, even if every app uses the same version of focus-trap, we will have multiple "separated and isolated" instances of focus-trap code in every app.

Yes, I realize that. In my view, this remains a bundling issue. A library should never bundle another library's code. Otherwise, it inevitably leads to issues like this.

Just like you would never bundle Vue or React or Angular for the same reasons. You only bundle that once for the app that will run in the browser. Otherwise, you declare a peer dependency on it with the version range you support.

> @stefcameron the issue is not with Nextcloud. If two apps built in different environments both ships their usage of `focus-trap`, they will face the same problem.

But are you running both apps in the same browser page (in the same DOM, not separate iframes since that shouldn't cause an issue with conflicting instances) at the same time for some reason?

> You can ignore `@nextcloud/focus-trap`, this is a try by us to mitigate the issue. We initially just used `focus-trap` on our apps and components.

👍

stefcameron:
> Multiple apps can be loaded on the same page. e.g
> 
> * the files app is one
> * the file viewer (unified file viewing Modal) is another one (opens when you click on a supported file)
> * the sidebar is another one (independent views and data)
> 
> On a huge ecosystem, you can have apps being enabled or not, loaded or not. They are all bundling their own libraries. Some might have different versions

OK, so you are working in some type of service where you have multiple bundled apps loaded in the same DOM. So you also have multiple versions of Vue and whatever else, because they're bundled independently.

So now, what is the specific issue you're running into with focus-trap? Why do you feel it's necessary to expose the global state to the window? What advantage is this giving you, or what problem is it solving?

stefcameron:
@skjnldsv I appreciate your continued patience with me as I try to understand this... 😄 

> We are trying to make traps stop fighting between each others, so having a queue against all traps on the same page instead of all traps in the same bundle scope is what we're trying to achieve. Otherwise we get infinite loops on the focus-trap library level since we have multiple runtime fighting to get the main focus :)

How is it that when one trap in one app is active, you're able to do _something_ (would also help to know what this _something_ is...) that causes a trap in a _different_ app to become active at the same time? The very concept of a trap is to trap the focus... I can think of a scenario where you might have a dialog which traps focus and has a button that triggers another dialog to open from it (not the best UX, IMO, but I've seen it), which then also has a trap that activates, thereby causing the trap in the parent dialog to be paused automatically. But would that inner dialog be a completely separate app instance? How do you end up with all these traps competing for focus?

skjnldsv:
> How do you end up with all these traps competing for focus?

1. Have a Modal opening
2. Have a menu inside that modal
3. Open the menu (dropdown), the menu takes the focus over the modal
4. See them fight for focus

docs/js/custom-active-focus-traps.js
```diff
@@ -0,0 +1,25 @@
...
+  const container = document.getElementById('custom-active-focus-traps');
+
+  const activeFocusTraps = {
+    activateTrap: (trap) => {
+      container.classList.add('is-active');
+    },
+    deactivateTrap: (trap) => {
+      container.classList.remove('is-active');
+    },
+  };
```
stefcameron, line 13: @skjnldsv In the override here, you're basically doing nothing, which means you're _preventing_ the trap that's already active from being paused when another one is being activated (or unpaused), and you're _preventing_ a previously-activate-and-now-paused trap from being unpaused when a trap gets deactivated.

Is this your real-world scenario, or just for the sake of the test?

If it's your real-world scenario (i.e. what you're actually doing in your override), how are you still not ending up with traps competing for focus? You still have 2 traps that have each added their own listeners to the document (via their own instances of `addListeners()`) and are both listening for focus events and wanting to pull focus into one another, away from the other one.

stefcameron: This must just be for the test, and it's not demonstrating anything meaningful in terms of why the feature exists. You should at least be creating 2 traps in here and doing something a bit more meaningful.

I'd still like to know what you're doing in your override that's actually preventing the two instances of focus-trap from fighting for focus so I can get a clearer mental picture.

I don't want to expose the guts of the whole internal state if possible. I'd like to find a better, more flexible solution that would be easier for me to maintain into the future.

vinicius73: Think about 2 isolated bundles, each one with this piece of code:

```ts
import { createFocusTrap } from 'focus-trap'

// --
```

We will have `app-a.js` and `app-b.js` 

Now think about @skjnldsv example.

Each bundle have an isolated instance of the function `createFocusTrap` and inside this bundle an isolate reference of `activeFocusTraps`

The object `activeFocusTraps` is a `closure` with a private member/state `trapQueue`
Now we have 2 `trapQueue` isolated from each other.

When `app-a.js` create an instance of `focus-trap` and active it, the `trapQueue` will receive it.
If `app-a.js` create another instance of `focus-trap`, everything will work.

But, if `app-b.js` creates an instance of `focus-trap` its state (`trapQueue`) do not know about `app-a.js` instance, so when `app-b.js` activate its instance while the instance of `app-a.js` is enabled, we will face an infinite loop.

That is exactly what is happening in our ecosystem.

stefcameron: @vinicius73 Yes, thank you, I understand all that. I've already worked out how you end-up in the conflict.

I think I can guess what you're doing with this new option, and that's to override both focus-trap instances and point both to the same implementation where you replicate the queue yourselves.

I will think about this some more.

stefcameron:
I did some thinking. I understand the use case, and I'm open to supporting it, but I'd like to do it in a way that exposes/replaces the _least_ of focus-trap internals as possible. As it stands, your current proposal here is to add an option that entirely replaces the internal stack, including how it works. I would like to reduce that to simply providing an option to give focus-trap an array that it should use as its stack, leaving the rest of the logic to focus-trap where it belongs.

So the option would be named `trapStack: Array<FocusTrap>` (because it's really a stack, not a queue) and you could provide your own array to it, which would be a global array in your case (you'd give the same one to all focus-trap instances).

vinicius73:
> So the option would be named `trapStack: Array<FocusTrap>` (because it's really a stack, not a queue) and you could provide your own array to it, which would be a global array in your case (you'd give the same one to all focus-trap instances).

That sounds very good @stefcameron 
Should I drop the PR and open another, or wait for you?

stefcameron:
@vinicius73 
> > So the option would be named `trapStack: Array<FocusTrap>` (because it's really a stack, not a queue) and you could provide your own array to it, which would be a global array in your case (you'd give the same one to all focus-trap instances).
> 
> That sounds very good @stefcameron Should I drop the PR and open another, or wait for you?

Wonderful! You can either repurpose this one, or open a new one and reference this one. I don't want to lose sight of our discussion (for posterity)! 🙂 

For the test, it should somehow make sure that the custom array gets used. I guess it can set one up in the `window` so that it can be accessed in the Cypress context.

vinicius73:
Closed in favor of #826

stefcameron:
@all-contributors add @nickvergessen for review

stefcameron:
@all-contributors add @vinicius73, @skjnldsv for ideas

stefcameron:
@all-contributors
add @vinicius73 for ideas
add @skjnldsv for ideas

stefcameron:
One last time...

@all-contributors add @skjnldsv for ideas

PR #932, opened 2023-03-30
Ensure focus is kept inside trap when focussed element is removed

kasperg:
focus-trap/focus-trap-react#962 has shown that if the element that is currently focused is removed from the DOM, the  focus is moved either to the document's body element or becomes `null`

When using `focus-trap` it would be appropriate to keep focus inside the trap. This is relevant if the trap is used for a modal and the content of the modal is updated but the modal stays open.

This adds an initial Cypress test to show whether this works or not.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>


docs/index.html
```diff
@@ -1135,6 +1135,34 @@ <h2 id="inert-heading">works with inert elements</h2>
...
+          This button will be removed when clicking it: <button id="dom-remove-button">remove</button>
+        </p>
+        <p>
+          <button id="deactivate-dom-remove" aria-describedby="dom-remove-heading">
+            deactivate trap
+          </button>
+        </p>
+      </div>
+      <p>
+        <a href="https://github.com/focus-trap/focus-trap/blob/master/docs/js/default.js">View demo source <span aria-hidden="true">&gt;&gt;</span></a>
```
stefcameron, line 1162: `default.js` should be `dom-remove.js` here

docs/index.html
```diff
@@ -1135,6 +1135,34 @@ <h2 id="inert-heading">works with inert elements</h2>
       </p>
     </div>
 
+
+    <div id="demo-dom-remove">
+      <h2 id="dom-remove-heading">Pulls focus when removing DOM elements</h2>
+      <p>
+      </p>
```
stefcameron, line 1142: I don't think the title really explains it well. How about adding this explanation here: "Keeps focus in the trap when the focused element is removed from the DOM."

stefcameron (changes requested):
Thank you for adding this handy demo and test (and for following nearby code patterns)!

docs/index.html
kasperg: :+1: Fixed in [c8756d7](https://github.com/focus-trap/focus-trap/pull/932/commits/c8756d74df588f37055740df7f0c536d1a2c90cc)

docs/index.html
kasperg: Agreed. I tried to give it another go in [46bf5cb](https://github.com/focus-trap/focus-trap/pull/932/commits/46bf5cb5c7309b8abd078a9a5da596f28393e73a).

stefcameron (commented):
Thanks for the updates so far.

kasperg:
@stefcameron I gave it a go implementing a fix. Tests are green so at least that's a start.

If you agree that this looks right I rewrite the commits a bit.

stefcameron:
@kasperg 
> @stefcameron I gave it a go implementing a fix. Tests are green so at least that's a start.
> 
> If you agree that this looks right I rewrite the commits a bit.

Thank you for the additional work you've done! I will check this out sometime over the next couple of days when I find a moment.

stefcameron (commented):
Thank you for your work on this! A couple of comments...

index.js
```diff
@@ -632,6 +651,16 @@ const createFocusTrap = function (elements, userOptions) {
       passive: false,
     });
 
+    if ('MutationObserver' in window) {
```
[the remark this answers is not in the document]
kasperg: I am not that much into server-side JS, but the check seems reasonable. I have added it in [409d9dc](https://github.com/focus-trap/focus-trap/pull/932/commits/409d9dc3b027fe884ad69fa5c4966fc5633c6965).

index.js
```diff
@@ -254,8 +258,8 @@ const createFocusTrap = function (elements, userOptions) {
       return false;
     }
 
-    if (node === undefined) {
-      // option not specified: use fallback options
+    if (node === undefined || !node.isConnected) {
```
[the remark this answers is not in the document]
kasperg: :+1: I have updated the the to use `isFocusable()` in [7f8b500](https://github.com/focus-trap/focus-trap/pull/932/commits/7f8b5001a73f1a1f908a8e2b5b266bfb51ff8434).

stefcameron (approved):
Looks great, thank you!

stefcameron (changes requested):
Actually, I think there's a gap in the case where the container elements are updated. The MutationObserver will keep observing old containers, no longer the right ones.

Specially related to this which I mentioned on the issue:

> As for the observer, I think there are two possible approaches there:
> 
> 1. It observes each trap container for node removals therein. Any time `updateContainerElements` is called, the observer is reset and restarts observing the new containers given. The callback checks the removed node against `state.mostRecentlyFocusedNode`.
> 2. It observes only the `state.mostRecentlyFocusedNode` and would be updated in `tryFocus()` and `checkFocusIn()` where that property is updated (or [the property](https://github.com/focus-trap/focus-trap/blob/master/index.js#L152) is converted into a `get()/set()` pair and the observer is updated in the setter so it doesn't matter where it gets updated).
> 
> I'm not sure how the observer reacts when the very node it's watching gets removed. Approach (1) might be the only way.

kasperg:
@stefcameron You are right. I totally missed the part about `updateContainerElements()`.

I updated the implementation with a simple change to fix address this in [1995abb](https://github.com/focus-trap/focus-trap/pull/932/commits/1995abb5993d647ade23430f5811401311304f3a).

However the code duplication lead me to reconsider the approach so I refactored most of it in [a832945](https://github.com/focus-trap/focus-trap/pull/932/commits/a832945456f439cc6bcd472256417e9953d92c65). Please check the commit comment for additional commentary.

If you agree with this new approach we can take it from there. If not we can continue from `1995abb`.

The PR had merge conflicts and needed a bit of cleanup so I have rebased the PR on `master`.

docs/js/index.js
```diff
@@ -59,3 +59,5 @@ require('./with-open-web-component')();
 // TEST MANUALLY (cypress-plugin-tab doesn't support inert)
 // http://localhost:9966/#demo-inert
 require('./inert')();
+
+require('./dom-remove')();
```
stefcameron, line 63: Just a tiny request to move this up to line 26 now that you've rebased. I reorganized the requires to put all the ones that are Cypress-testable before all those that aren't, just to make it easier for me to follow the "list" when doing manual testing.

stefcameron (commented):
@kasperg I like your new approach to solving the problem. Seems very nice. Your thought process/explanation for changing directions a bit was very helpful. I just have one final small request, which I think I'll just push up to your branch myself since it's just a line in the `/docs/index.js` now that you've rebased.

stefcameron (approved):
🚀

stefcameron:
If you're good with that, I'll merge it. 😄

kasperg:
👍 Thanks. Appreciate the help.

stefcameron:
@all-contributors add @kasperg for bug, code

stefcameron:
Published in focus-trap v7.4.3

PR #940, opened 2023-04-10
fix(typings): Updates typings for the document option to support shadow DOM host elements

### LINKED ISSUE #939: Update type for document option to allow for shadowRoot elements.
According to the docs for shadow DOM (https://github.com/focus-trap/focus-trap#shadow-dom) 

> You must use the document option to configure the focus trap to use your shadow host element as its document. The downside of this option is that, while selector queries on nodes inside your trap will now work, the trap will not prevent focus from being set on nodes outside your Shadow DOM, which is the same drawback as putting a focus trap [inside an iframe](https://focus-trap.github.io/focus-trap/#demo-in-iframe).

If I'm reading this correctly, for elements with shadowRoot, I should be able to set the document to my host element?

However, the typings for the `document` property  only allow for the `Document`. Should this type be expanded to include any HTMLElement?

https://github.com/focus-trap/focus-trap/blob/328022bf2f0fa44fa0bcb210ab9ca09198a20b52/index.d.ts#L186-L190

driskull:
Modifies the typescript typings file to add to the document option so that a typescript user can set any HTML element as the document. This will allow the document to be set to shadow DOM elements.

Fixes #939

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -185,9 +185,9 @@ declare module 'focus-trap' {
     delayInitialFocus?: boolean;
     /**
      * Default: `window.document`. Document where the focus trap will be active.
-     * This allows to use FocusTrap in an iFrame context.
+     * This allows to use FocusTrap in an iFrame context or to configure the focus trap to use your shadow host element as its document.
      */
-    document?: Document;
+    document?: Document | HTMLElement;
 
     /**
      * Specific tabbable options configurable on focus-trap.
--- a/.changeset/soft-actors-cheat.md
+++ b/.changeset/soft-actors-cheat.md
@@ -1,0 +1,5 @@
+---
+'focus-trap': patch
+---
+
+Updates typings for the document option to support shadow DOM host elements


index.d.ts
```diff
@@ -185,9 +185,9 @@ declare module 'focus-trap' {
     delayInitialFocus?: boolean;
     /**
      * Default: `window.document`. Document where the focus trap will be active.
-     * This allows to use FocusTrap in an iFrame context.
+     * This allows to use FocusTrap in an iFrame context or to configure the focus trap to use your shadow host element as its document.
      */
-    document?: Document;
+    document?: Document | HTMLElement;
```
stefcameron, line 190: The only other option would be https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot because of the expectation to use the `.activeElement` API. But I don't think this is necessary. See my comment on the related issue you opened earlier.

stefcameron (changes requested):
Thank you for your proactiveness in suggesting a PR. I think it's a docs issue, however...

PR #960, opened 2023-05-12
fix: Clicking on open shadowDOM components within a focus trap's container when `clickOutsideDeactivates=true` should not deactivate the focus trap

### LINKED ISSUE #959: Clicking another shadowed element inside the focus trap will deactivate the focus trap
When the following settings are set, clicking another shadowed element inside the focus trap will deactivate the focus trap when it shouldn't.

```ts
clickOutsideDeactivates: true,
escapeDeactivates: false,
```

bug

Clicking the "Light Dom Button" was just to verify a normal button doesn't deactivate the focus trap. However, a custom element button with a shadowRoot does deactivate the focus trap.

## Example

Example demo patch to see problem

```patch
diff --git forkSrcPrefix/docs/js/in-open-shadow-dom.js forkDstPrefix/docs/js/in-open-shadow-dom.js
index 0377780fad82edafdefa7b9cd02f444042a6f267..e3273d9bb82b641cf019bdb1b9e7e9c1a79c445d 100644
--- forkSrcPrefix/docs/js/in-open-shadow-dom.js
+++ forkDstPrefix/docs/js/in-open-shadow-dom.js
@@ -1,5 +1,13 @@
 const { createFocusTrap } = require('../../index');
 module.exports = () => {
+  class CustomButton extends HTMLElement {
+    constructor() {
+      super();
+
+      this.attachShadow({ mode: 'open' }).innerHTML = '<button><slot></slot></button>';
+    }
+  }
+
   class FocusTrapModal extends HTMLElement {
     constructor() {
       super();
@@ -14,6 +22,8 @@ module.exports = () => {
           <a href="#">with</a> <a href="#">some</a> <a href="#">focusable</a> parts.
         </p>
         <p>
+          <custom-button>Shadow Button</custom-button>
+          <button>Light DOM Button</button>
           <button id="deactivate-in-open-shadow-dom" aria-describedby="in-open-shadow-dom-heading">
             deactivate trap
           </button>
@@ -32,7 +42,8 @@ module.exports = () => {
       const focusTrap = createFocusTrap(modalEl, {
         onActivate: () => modalEl.classList.add('is-active'),
         onDeactivate: () => modalEl.classList.remove('is-active'),
-        escapeDeactivates: true,
+        clickOutsideDeactivates: true,
+        escapeDeactivates: false,
       });
 
       document
@@ -45,4 +56,5 @@ module.exports = () => {
   }
 
   customElements.define('focus-trap-modal', FocusTrapModal);
+  customElements.define('custom-button', CustomButton);
 };

```

example commit: https://github.com/driskull/focus-trap/commit/1be26a4154441623d52a068b756855235f26ad93

I tracked the issue down to this function

https://github.com/focus-trap/focus-trap/blob/b781078e26c5a1b0d0b34250d361ee440bd3fe2e/index.js#L185-L196

It seems to be having a problem with the slotted `<slot>` element.

This seems to fix it but I'm not sure if its the right solution...

```ts
tabbableNodes.find(function (node) {
  return node === element || node.contains(element);
});
```

I did notice that even elements that are not "tabbable" in this scenario are deactivating the focus trap. https://github.com/driskull/focus-trap/commit/ade43e5a33f23e79b08608d378deed332c68f9b6

Let me know how I can help with this issue. :)

driskull:
## Summary

- Adds logic in `findContainerIndex()` to optionally see if the `event.composedPath` contains a state container. If it does, it will not deactivate the focus trap. Fixes #959
- Adds comment note to manually test the example.
- Modifies an example to demonstrate the issue when clicking on open shadowDOM components within a focus trap container element. 

## Repro

Demo: `/focus-trap/focus-trap/blob/master/docs/js/in-open-shadow-dom.js`

- Clicking on the `<custom-span>` element shouldn't deactivate the focus trap.
  - This custom element has no tabbable elements within it and it is not tabbable either.
- Clicking on the `<custom-button>` element shouldn't deactivate the focus trap.
  - This custom element has a `<button>` inside which is picked up by `tabbable` as can be seen by looking at the focus trap `containerGroups` but it still deactivates the focus trap when clicked.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>


index.js
```diff
@@ -379,8 +379,10 @@ const createFocusTrap = function (elements, userOptions) {
   // so that it precedes the focus event.
   const checkPointerDown = function (e) {
     const target = getActualTarget(e);
+    const composedPath = typeof e.composedPath === 'function' && e.composedPath();
+    const clickedInsideContainer = composedPath && state.containers.some(container => composedPath.includes(container));
 
-    if (findContainerIndex(target) >= 0) {
+    if (findContainerIndex(target) >= 0 || clickedInsideContainer) {
```
driskull, line 385: I'm wondering if we need both of these checks?

stefcameron: Actually, the `composedPath()` check should be _inside_ `findContainerIndex()` so that the index is returned if the `target` is indeed inside one of the containers. That way, we cover other places where `findContainerIndex()` is called for the same purpose.

stefcameron (changes requested):
PR looks great except for where you're currently checking the composed path...

driskull:
@stefcameron checked in updates. The only question is that the current findContainerIndex function only takes a target so we'd have to refactor to take an event or composedPath as well.

Let me know what you think.

stefcameron:
@driskull 
> @stefcameron checked in updates. The only question is that the current findContainerIndex function only takes a target so we'd have to refactor to take an event or composedPath as well.
> 
> Let me know what you think.

Yeah, that's not as elegant as I had pictured it in my head, but I think what you've done there is the best we can do. Thanks for working through it.

Just need to pass the `composedPath` into that function via `checkKeyNav()`, `checkClick()` and `checkFocusIn()` as well to cover all the bases (although key-nav and focus-in imply the target is tabbable, which means the composed path shouldn't be necessary to determine it's in a container, but for completeness, and in case there's some odd corner case, it seems better to just do it in these two spots as well).

index.js
driskull: Done ✅

driskull:
Updated PR with suggestions. 👍

stefcameron (changes requested):
Nice update. It was looking great... until I went through the (unfortunately required) manual testing list in `/docs/js/index.js` and discovered that the "working inside an iframe" (`in-iframe.js`) demo is broken: While you're expected to be able to click on elements _outside_ the iframe even when the trap is active, you're not expected to be able to click on the "something else" button that's _inside_ the iframe, but outside the trap's container.

The reason this happens is because of our logic, so far, in `findContainerIndex()` WRT to the `composedPath` fallback. Even in the iframe case, the `composedPath` for the "something else" button goes all the way back up to the iframe's window, and our check isn't exclusive enough.

Basically, I think the check will always succeed because it's not being restricted to the `container` being checked.

We need to look in the composed path for `element`, but no "higher" (up the stack) than the `container` itself, which is what the `Node.contains()` API does (I checked that `container.contains(container) === true`).

Since `composedPath` is a path from the element up to the iframe's window, when clicking on the "something else" button outside the trap, it looks like this:

image

When clicking on the "nothing" button (inside the trap), it looks like this:

image

All that to say, I think our check needs to be this:

```javascript
composedPath?.includes(container)
```

WDYT?

driskull:
@stefcameron yes that makes perfect sense. 

I'll also update to use the optional chaining. I'm so used to just writing TS I forget that it's es6.

I'll run through the iframe tests as well and see if it fixes it before checking in.

driskull:
iframe test looks good now. Same with the other shadowed component tests.

index.js
```diff
@@ -193,7 +195,8 @@ const createFocusTrap = function (elements, userOptions) {
         //  web components if the `tabbableOptions.getShadowRoot` option was used for
         //  the trap, enabling shadow DOM support in tabbable (`Node.contains()` doesn't
         //  look inside web components even if open)
-        tabbableNodes.find((node) => node === element)
+        tabbableNodes.find((node) => node === element) ||
+        composedPath?.includes(container)
```
driskull, line 199: Should the `includes` test be first since it may be faster than the `find`?

stefcameron: Yes, actually. In our very initial approach, this order would've been better, but thanks to your `composedPath` idea, I think you're right. The composed path test should be second, and then `find()` be third.

driskull: done!

stefcameron (commented):
Thanks for the final adjustment! The last thing to do is to include an updated demo bundle (`npm run demo-bundle` and commit the result) __after__ you're done running Cypress tests because running those tests also generates the bundle, but does so with the wrong runtime flags resulting in the "inside iframe" demo not working on the demo page. Something I need to figure out how to prevent.

driskull:
@stefcameron should I create a separate test for our use case? Because I added the `clickOutsideDeactivates: true,` the existing test fails when testing that the click should be prevented.

driskull:
I think we just need some UI to toggle `clickOutsideDeactivates: true` for the manual testing case. I'll see if I can do that.

docs/index.html
stefcameron, line 1: I'm wondering what suddenly happened to the HTML file. It's like Prettier was run against it or some type of auto-formatting took place. focus-trap's Prettier settings wouldn't have made all these changes, however.

docs/js/in-open-shadow-dom.js
```diff
@@ -29,20 +50,50 @@ module.exports = () => {
...
+
+      const checkbox = document.getElementById(
+        'checkbox-clickoutsidedeactivates-in-open-shadow'
+      );
+
+      checkbox.addEventListener('change', function () {
+        clickOutsideDeactivates = checkbox.checked;
+        focusTrap.deactivate();
+        focusTrap = initialize();
       });
```
stefcameron, line 77: The addition of the checkbox to control the outside click behavior is nice, but I don't think it was totally necessary. We could've just kept the demo with that setting added in, and a comment next to the setting explaining why it was important to have that setting turned on. With the checkbox there now, I feel like it becomes important to describe, in the demo, its reason for being there (especially since we have to manually test this one too) -- that being that even though an outside click will disable, clicking particularly on the "shadow span" inside the trap will not be considered outside the trap, and thus the trap will remain active.

driskull:
reverted those changes and commented out the `clickOutsideDeactivates`

docs/index.html
driskull: fixed.

stefcameron (approved):
This looks solid, thank you! 🚀

stefcameron:
@all-contributors add @driskull for bug, code, tutorial

PR #974, opened 2023-06-01
Support elements with positive tabindex attributes in single-container traps

### LINKED ISSUE #375: focus-trap stuck on elements with positive tabIndex

stefcameron:
Fixes #375

ALL existing cypress tests are passing.

The goal is to take the least invasive approach when it comes to enforcing tab order. Let the browser do as much of the work as possible, and only interject when it doesn't do what we expect it to (which is the general theme of focus-trap).

Here's the working demo, including negative tabindex edge case support:

focus-trap-positive-tabindex

DONE:

- [x] artificially limit the trap to a single container if at least one positive tabindex node is discovered (check for it in `updateTabbableNodes()`
    - `if >1 container && >0 positive tabindex nodes found, then throw Error("trap not supported")`
- [x] handle edge case where MRU focused node had negative tabindex
    - in this case, the browser apparently sets focus on the next element in DOM order, not tabindex order
    - or just ignore this as a rare edge case
- [x] get rid of DEBUG comments
- [x] will need to share logic from tabbable about determining tabindex somehow
  - Keep this Tabbable commit except for the multi-container-related typing updates: https://github.com/focus-trap/tabbable/commit/5b753b6f8d190a60ff636553a83d0eaae8d58492
- [x] check manually tested examples
- [x] see if a cypress test can be added for the new positive-tabindex example
- [x] add changeset

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>


DaviDevMod:
I made a few adjustments in [my fork](https://github.com/DaviDevMod/focus-trap-not-mine/tree/positive-tabindex).
I'm going to reproduce the changes here.

By the way, I noticed this:

https://github.com/focus-trap/focus-trap/blob/f22e3611a6eebbf1245f34e0a37a86250a5875fc/index.js#L72

Which can be removed, since *tabbable* doesn't support IE anymore.
But this change doesn't belong to this PR.

DaviDevMod:
@stefcameron It looks like I can't review arbitrary lines, eg, I can't review line 516 of index.js to make [this suggestion](https://github.com/DaviDevMod/focus-trap-not-mine/commit/b5a5f191a02f30e20026bef454076b9942e08a2e).

So either I open another PR against `focus-trap/positive-tabindex` or you pull the changes from my fork.

stefcameron:
@DaviDevMod I think it's because using `suggestion` in a comment equates to writing directly to a branch. Looks like you made 3 commits in your fork. Might be better to make a PR into this `positive-tabindex` branch. I should then be able to merge that into this branch/PR, and then the whole into `master`.

Unless it's just that one suggestion you're wanting to make, in which case I can easily add that myself, but just wondering what's going on there.

stefcameron:
@all-contributors add @DaviDevMod for code, bug

PR #987, opened 2023-06-27
Positive tabindex

DaviDevMod:
This PR is meant to fix a couple of bugs in https://github.com/focus-trap/focus-trap/pull/974 and, at the same time, simplify some logic around the fixes.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/index.js
+++ b/index.js
@@ -366,7 +366,15 @@ const createFocusTrap = function (elements, userOptions) {
             //  regardless of positive tabindexes, if any -- and for reasons explained in the NOTE
             //  above related to `firstDomTabbable` and `lastDomTabbable` properties, we fall back to
             //  basic DOM order
-            return forward ? firstDomTabbableNode : lastDomTabbableNode;
+            if (forward) {
+              return focusableNodes
+                .slice(focusableNodes.indexOf(node) + 1)
+                .find((el) => isTabbable(el));
+            }
+
+            return focusableNodes
+              .slice(0, focusableNodes.indexOf(node))
+              .findLast((el) => isTabbable(el));
           }
 
           return tabbableNodes[nodeIdx + (forward ? 1 : -1)];
--- a/index.js
+++ b/index.js
@@ -358,7 +358,7 @@ const createFocusTrap = function (elements, userOptions) {
          * @returns {HTMLElement|undefined} The next tabbable node, if any.
          */
         nextTabbableNode(node, forward = true) {
-          const nodeIdx = tabbableNodes.findIndex((n) => n === node);
+          const nodeIdx = tabbableNodes.indexOf(node);
           if (nodeIdx < 0) {
             // either not tabbable nor focusable, or was focused but not tabbable (negative tabindex):
             //  since `node` should at least have been focusable, we assume that's the case and mimic
@@ -369,15 +369,7 @@ const createFocusTrap = function (elements, userOptions) {
             return forward ? firstDomTabbableNode : lastDomTabbableNode;
           }
 
-          if (forward) {
-            // first found, if any (array will be empty if `node` was in last position)
-            return tabbableNodes.slice(nodeIdx + 1).find(() => true);
-          }
-
-          return tabbableNodes
-            .slice(0, nodeIdx) // exclude `node` itself
-            .reverse()
-            .find(() => true); // first found, if any (array will be empty if `node` was in first position)
+          return tabbableNodes[nodeIdx + (forward ? 1 : -1)];
         },
       };
     });
--- a/index.js
+++ b/index.js
@@ -358,18 +358,26 @@ const createFocusTrap = function (elements, userOptions) {
          * @returns {HTMLElement|undefined} The next tabbable node, if any.
          */
         nextTabbableNode(node, forward = true) {
-          // Can't use `tabbableNodes` because `node` may have negative tab index.
-          const nodeIdx = focusableNodes.indexOf(node);
+          const nodeIdx = tabbableNodes.indexOf(node);
+          if (nodeIdx < 0) {
+            // either not tabbable nor focusable, or was focused but not tabbable (negative tabindex):
+            //  since `node` should at least have been focusable, we assume that's the case and mimic
+            //  what browsers do, which is set focus to the next node in __document position order__,
+            //  regardless of positive tabindexes, if any -- and for reasons explained in the NOTE
+            //  above related to `firstDomTabbable` and `lastDomTabbable` properties, we fall back to
+            //  basic DOM order
+            if (forward) {
+              return focusableNodes
+                .slice(focusableNodes.indexOf(node) + 1)
+                .find((el) => isTabbable(el));
+            }
 
-          if (forward) {
             return focusableNodes
-              .slice(nodeIdx + 1)
-              .find((el) => isTabbable(el));
+              .slice(0, focusableNodes.indexOf(node))
+              .findLast((el) => isTabbable(el));
           }
 
-          return focusableNodes
-            .slice(0, nodeIdx)
-            .findLast((el) => isTabbable(el));
+          return tabbableNodes[nodeIdx + (forward ? 1 : -1)];
         },
       };
     });
--- a/index.js
+++ b/index.js
@@ -366,14 +366,16 @@ const createFocusTrap = function (elements, userOptions) {
             //  regardless of positive tabindexes, if any -- and for reasons explained in the NOTE
             //  above related to `firstDomTabbable` and `lastDomTabbable` properties, we fall back to
             //  basic DOM order
+            const nodeFocusableIndex = focusableNodes.indexOf(node);
+
             if (forward) {
               return focusableNodes
-                .slice(focusableNodes.indexOf(node) + 1)
+                .slice(nodeFocusableIndex + 1)
                 .find((el) => isTabbable(el));
             }
 
             return focusableNodes
-              .slice(0, focusableNodes.indexOf(node))
+              .slice(0, nodeFocusableIndex)
               .findLast((el) => isTabbable(el));
           }
 
--- a/index.js
+++ b/index.js
@@ -358,28 +358,18 @@ const createFocusTrap = function (elements, userOptions) {
          * @returns {HTMLElement|undefined} The next tabbable node, if any.
          */
         nextTabbableNode(node, forward = true) {
-          const nodeIdx = tabbableNodes.indexOf(node);
-          if (nodeIdx < 0) {
-            // either not tabbable nor focusable, or was focused but not tabbable (negative tabindex):
-            //  since `node` should at least have been focusable, we assume that's the case and mimic
-            //  what browsers do, which is set focus to the next node in __document position order__,
-            //  regardless of positive tabindexes, if any -- and for reasons explained in the NOTE
-            //  above related to `firstDomTabbable` and `lastDomTabbable` properties, we fall back to
-            //  basic DOM order
-            const nodeFocusableIndex = focusableNodes.indexOf(node);
+          // Can't use `tabbableNodes` because `node` may have negative tab index.
+          const nodeIdx = focusableNodes.indexOf(node);
 
-            if (forward) {
-              return focusableNodes
-                .slice(nodeFocusableIndex + 1)
-                .find((el) => isTabbable(el));
-            }
-
+          if (forward) {
             return focusableNodes
-              .slice(0, nodeFocusableIndex)
-              .findLast((el) => isTabbable(el));
+              .slice(nodeIdx + 1)
+              .find((el) => isTabbable(el));
           }
 
-          return tabbableNodes[nodeIdx + (forward ? 1 : -1)];
+          return focusableNodes
+            .slice(0, nodeIdx)
+            .findLast((el) => isTabbable(el));
         },
       };
     });
--- a/index.js
+++ b/index.js
@@ -466,7 +466,7 @@ const createFocusTrap = function (elements, userOptions) {
       const containerGroup =
         containerIndex >= 0 ? state.containerGroups[containerIndex] : undefined;
 
-      if (!containerGroup) {
+      if (containerIndex < 0) {
         // target not found in any group: quite possible focus has escaped the trap,
         //  so bring it back into...
         if (isBackward) {
@@ -481,19 +481,36 @@ const createFocusTrap = function (elements, userOptions) {
       } else if (isBackward) {
         // REVERSE
 
+        // is the target the first tabbable node in a group?
+        let startOfGroupIndex = findIndex(
+          state.tabbableGroups,
+          ({ firstTabbableNode }) => target === firstTabbableNode
+        );
+
         if (
-          target === containerGroup.container ||
-          containerGroup.focusableNodes.indexOf(target) <=
-            containerGroup.focusableNodes.indexOf(
-              containerGroup.firstTabbableNode
-            )
+          startOfGroupIndex < 0 &&
+          (containerGroup.container === target ||
+            (isFocusable(target, config.tabbableOptions) &&
+              !isTabbable(target, config.tabbableOptions) &&
+              !containerGroup.nextTabbableNode(target, false)))
         ) {
+          // an exception case where the target is either the container itself, or
+          //  a non-tabbable node that was given focus (i.e. tabindex is negative
+          //  and user clicked on it or node was programmatically given focus)
+          //  and is not followed by any other tabbable node, in which
+          //  case, we should handle shift+tab as if focus were on the container's
+          //  first tabbable node, and go to the last tabbable node of the LAST group
+          startOfGroupIndex = containerIndex;
+        }
+
+        if (startOfGroupIndex >= 0) {
           // YES: then shift+tab should go to the last tabbable node in the
           //  previous group (and wrap around to the last tabbable node of
           //  the LAST group if it's the first tabbable node of the FIRST group)
           const destinationGroupIndex =
-            (containerIndex - 1 + state.containerGroups.length) %
-            state.containerGroups.length;
+            startOfGroupIndex === 0
+              ? state.tabbableGroups.length - 1
+              : startOfGroupIndex - 1;
 
           const destinationGroup = state.tabbableGroups[destinationGroupIndex];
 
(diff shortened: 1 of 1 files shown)


DaviDevMod:
Just noticed that the description for the commit 91da5cac650e3fc141d8990f15838337475c67b1 is wrong:

> checking that `target` is either the `lastTabbableNode` in the container or a focusable (and not tabbable) node **PRECEDING** it in document order.

Should have been: "[...] or a focusable (and not tabbable) node **FOLLOWING** it in document order."

There's also a symmetric condition (in case they are tabbing backward) with `firstTabbableNode` rather than `lastTabbableNode` and in that case the condition is indeed checking that  `target` is either the `firstTabbableNode` in the container or a focusable (and not tabbable) node **PRECEDING** it in document order.

index.js
```diff
@@ -358,26 +358,28 @@ const createFocusTrap = function (elements, userOptions) {
...
+            return focusableNodes
+              .slice(0, nodeFocusableIndex)
+              .findLast((el) => isTabbable(el));
           }
 
-          return tabbableNodes
-            .slice(0, nodeIdx) // exclude `node` itself
-            .reverse()
-            .find(() => true); // first found, if any (array will be empty if `node` was in first position)
+          return tabbableNodes[nodeIdx + (forward ? 1 : -1)];
```
stefcameron: This could lead to a positive or negative index that doesn't exist in the `tabbableNodes` array. Technically, the result should be `undefined`, but it feels a bit wrong to access an array out of bounds.

index.js
```diff
@@ -466,7 +468,7 @@ const createFocusTrap = function (elements, userOptions) {
       const containerGroup =
         containerIndex >= 0 ? state.containerGroups[containerIndex] : undefined;
 
-      if (containerIndex < 0) {
+      if (!containerGroup) {
```
stefcameron, line 461: The rest of the changes below this line are too much for me to grok. Feels like there's a lot of carefully-honed logic in there that we're deconstructing and changing around. Can you explain the changes in behavior happening in here? You're even using the modulus operator?! 😉 I'm also concerned about some key comments going away. I rely on comments to remember what was going on in some edge case and why we did something because I'm not in this on a daily basis. Would be good to retain those that are still relevant, or replace them with updated ones.

stefcameron (commented):
Thanks for these updates. I'm just having a hard time understanding what's going. Seems like a lot of changes...

index.js
DaviDevMod: As a person who learned to code with C, I feel you. But in JavaScript it's a safe operation.
In any case I pushed f198c2b right while you writing this comment :smile: and the out of bound access is gone already.

index.js
DaviDevMod: > The rest of the changes below this line are too much for me to grok.

Let's break it down into manageable pieces.
We have the block of this `if (!containerGroup)` condition you mentioned, which I din't touch, and then two `else` blocks: one for when the user is tabbing forward and one for when they are tabbing backward. The two blocks are quite symmetric, the logic is the same, so let's consider just the block for the case in which they are tabbing forward:

https://github.com/focus-trap/focus-trap/blob/f198c2bcbcb9b09e08b7bb3a75288d0163983880/index.js#L501

This is the whole block of code:
```js
if (
  target === containerGroup.container ||
  containerGroup.focusableNodes.indexOf(target) >=
    containerGroup.focusableNodes.indexOf(
      containerGroup.lastTabbableNode
    )
) {
  // YES: then tab should go to the first tabbable node in the next
  //  group (and wrap around to the first tabbable node of the FIRST
  //  group if it's the last tabbable node of the LAST group)
  const destinationGroupIndex =
    (containerIndex + 1) % state.containerGroups.length;

  const destinationGroup = state.tabbableGroups[destinationGroupIndex];

  destinationNode =
    getTabIndex(target) >= 0
      ? destinationGroup.firstTabbableNode
      : destinationGroup.firstDomTabbableNode;
} else if (!isTabEvent(event)) {
  // user must have customized the nav keys so we have to move focus manually _within_
  //  the active group: do this based on the order determined by tabbable()
  destinationNode = containerGroup.nextTabbableNode(target);
}
```

The condition:
```js
if (
  target === containerGroup.container ||
  containerGroup.focusableNodes.indexOf(target) >=
    containerGroup.focusableNodes.indexOf(
      containerGroup.lastTabbableNode
    )
```

is assessing whether `target` is itself the `container` containing `target`; or the `lastTabbableNode` in the `container`; or a focusable node within the `container` that follows `lastTabbableNode` in document order.
If `target` is one of these elements, focus-trap can't trust the browser, it has to intervene and programmatically pass the focus to an element in the next container.
That's how we get the index of the next container:

```js
const destinationGroupIndex =
  (containerIndex + 1) % state.containerGroups.length;
```
> You're even using the modulus operator?!  

It has been waiting its whole life for this moment :laughing: it's the perfect use case for it.

So now that we have the next container, what's the right element we need within it?
As we discussed in the looong issue we are trying to close, if `target` has a negative tab index, the focus can't simply go to the `firstTabbableNode` of the container, because that takes into account tab indexes (meaning it could be a positive tab index that is the last element in document order within the container) while browsers just pick the closest tabbable element in document order, without looking at tab index values. So the element to focus is:

```js
destinationNode =
  getTabIndex(target) >= 0
    ? destinationGroup.firstTabbableNode
    : destinationGroup.firstDomTabbableNode;
```

The last couple of lines in the block (that `else if`) were already there, and I didn't touch them.
Though I did touch the body of `nextTabbableNode`, which is being called there.

DaviDevMod: Actually I inherited the condition `target === containerGroup.container` from here:

https://github.com/focus-trap/focus-trap/blob/f22e3611a6eebbf1245f34e0a37a86250a5875fc/index.js#L533

But I notice only now that it's wrong. It is necessary only when tabbing backward, because the container itself would be like the very first focusable element in the container. But when tabbing forward, we don't care about it, because we only look at the last focusable elements.

I'm going to fix it.

DaviDevMod: Removing `target === containerGroup.container` from the case in which they are tabbing forward makes a test fail.
I'm not sure why, an hypothesis is that in that case `findNextNavNode` is not being called on `keydown` but rather on `focusin`:

https://github.com/focus-trap/focus-trap/blob/f198c2bcbcb9b09e08b7bb3a75288d0163983880/index.js#L441

I haven't yet considered the implications of this case, and right now I really need to get some sleep.
Will look into it as soon as I can.

index.js
stefcameron: I started with C++. 😉

stefcameron (commented):
I tried to read through everything as carefully as possible again. I greatly appreciate the time you took to write your explanation of everything. I get the logic, but my hesitation is two-fold:

1. My PR was specifically addressing handling nodes with positive tabindexes. That surfaced one edge case: Where to go when the node tabbed _from_ had a negative tabindex, and I did plug that whole. Your PR is trying to fix _all_ cases where it's technically possible to tab away from a negative tabindex node. Probably should _not_ be lumped into my PR.
2. In `findNextNavNode()`, you're changing a lot of carefully curated logic. 🙂 That's OK, but while I understand the changes you're making, I'm concerned about the rewrite of `nextTabbableNode()` to use only focusable nodes without considering tabbable nodes and the order given by Tabbable, and I'm concerned about all the edge cases we may no longer handle in `findNextNavNode()`. This is the really hard part for me. The code that's there now is simply trying to move focus to the first tabbable node in the next container in container order (not DOM container order, technically flawed as that may be) and wrap around if necessary. It's also trying to decide what to do if the node tabbed from wasn't tabbable or the container itself and doing a reset in that case. That's the specific piece I think you're trying to fix, but I'm concerned with all the other stuff moving around because of it.

I think first thing is that your PR should be separate from mine, after I merge mine into master, unless there's something specific with positive tabindex support that mine is failing to do.

stefcameron:
Also remember that the fact that the tests pass is not a 100% indication that all is good. There are unfortunately a number of tests that require manual testing because they can't be tested in Cypress. 😞

index.js
stefcameron: Not sure. That's why I'm so hesitant to make changes in there... `findNextNavNode()` is being called on `focusin` and `keydown`, but in `focusin` only in the case where you're tabbing across containers. It's always called on `keydown`.

I'm sure that clause is there for a reason.

index.js
```diff
@@ -358,26 +358,18 @@ const createFocusTrap = function (elements, userOptions) {
          * @returns {HTMLElement|undefined} The next tabbable node, if any.
          */
         nextTabbableNode(node, forward = true) {
```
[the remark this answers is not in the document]
DaviDevMod: Yes, you are right.

Still in the case for `nodeIdx < 0`

https://github.com/focus-trap/focus-trap/blob/f22e3611a6eebbf1245f34e0a37a86250a5875fc/index.js#L362

needs to be fixed, with these changes: https://github.com/focus-trap/focus-trap/pull/987/commits/3e56ae38983393dc9eb81aeb64d00eab21684fcb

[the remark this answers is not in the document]
DaviDevMod: Here I reverted the changes: [c81b506](https://github.com/focus-trap/focus-trap/pull/987/commits/c81b506c94bf22d8a5350b8319b467aee784b84b)

index.js
DaviDevMod: After reverting the changes: c81b506

 We got the out of bound access again: `return tabbableNodes[nodeIdx + (forward ? 1 : -1)];`
Even though it's not undefined behaviour in JS, I understand that it's not good looking, but neither is the code it is simplifying:

```js
if (forward) {
  // first found, if any (array will be empty if `node` was in last position)
  return tabbableNodes.slice(nodeIdx + 1).find(() => true);
}

return tabbableNodes
  .slice(0, nodeIdx) // exclude `node` itself
  .reverse()
  .find(() => true); // first found, if any (array will be empty if `node` was in first position)

```

At the end it's just a question of preference, the library would work either way, so if you prefer that way I revert the changes.

index.js
DaviDevMod: Alright, I haven't even looked at where `findNextNavNode` is called, I just trust that it's better to minimise the amount of changes.

But [replying to](https://github.com/focus-trap/focus-trap/pull/987#pullrequestreview-1503490260):

>Your PR is trying to fix all cases where it's technically possible to tab away from a negative tabindex node. Probably should not be lumped into my PR. 

I think that at least the changes in https://github.com/focus-trap/focus-trap/pull/987/commits/b5a5f191a02f30e20026bef454076b9942e08a2e should be introduced:

```js
destinationNode =
  getTabIndex(target) >= 0
    ? destinationGroup.firstTabbableNode
    : destinationGroup.firstDomTabbableNode;
```

because if positive tab indexes are not supported, `firstDomTabbableNode` coincides with `firstTabbableNode` , the distinction between the two should be something to be taken care of at the moment of introducing support for positive tab indexes.

DaviDevMod: Reverted changes in to `findNextNavNode` here: 20af8f2.
(Except for the changes [discussed above](https://github.com/focus-trap/focus-trap/pull/987#discussion_r1245715986): https://github.com/focus-trap/focus-trap/commit/b5a5f191a02f30e20026bef454076b9942e08a2e)

index.js
```diff
@@ -522,39 +501,25 @@ const createFocusTrap = function (elements, userOptions) {
...
           // YES: then tab should go to the first tabbable node in the next
           //  group (and wrap around to the first tabbable node of the FIRST
           //  group if it's the last tabbable node of the LAST group)
           const destinationGroupIndex =
-            lastOfGroupIndex === state.tabbableGroups.length - 1
-              ? 0
-              : lastOfGroupIndex + 1;
+            (containerIndex + 1) % state.containerGroups.length;
 
           const destinationGroup = state.tabbableGroups[destinationGroupIndex];
```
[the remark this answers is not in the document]
DaviDevMod: This was fixed in https://github.com/focus-trap/focus-trap/commit/20af8f2bcedadf5256ca3da8d01e55624cebc278.
See also the [discussion](https://github.com/focus-trap/focus-trap/pull/987#discussion_r1245722263) leading to the commit.

index.js
```diff
@@ -481,39 +473,26 @@ const createFocusTrap = function (elements, userOptions) {
       } else if (isBackward) {
         // REVERSE
 
-        // is the target the first tabbable node in a group?
-        let startOfGroupIndex = findIndex(
-          state.tabbableGroups,
-          ({ firstTabbableNode }) => target === firstTabbableNode
-        );
-
         if (
```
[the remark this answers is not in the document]
DaviDevMod: Do you mean to have a comment emphasising that we are not in the block of [this `if` statement](https://github.com/DaviDevMod/focus-trap-not-mine/blob/20af8f2bcedadf5256ca3da8d01e55624cebc278/index.js#L469)?

```js
if (containerIndex < 0) {
  ...
} else if (isBackward) {
  // `containerGroup` is granted to exist here
  ...
}
```

In any case, I think it's better to leave comments up to you :laughing:

index.js
DaviDevMod: Just realised that the https://github.com/focus-trap/focus-trap/commit/b5a5f191a02f30e20026bef454076b9942e08a2e won't do anything without the rest of of the changes, because now the destination node is set only [`if (startOfGroupIndex >= 0)`](https://github.com/focus-trap/focus-trap/blob/20af8f2bcedadf5256ca3da8d01e55624cebc278/index.js#L506C9-L506C37) which doesn't take into account focusable, and not tabbable, nodes.

I could fix just that, BUT... I think we can get rid of `focusin` altogether and reintroduce all the changes I was proposing.

`focusin` was needed before because the library couldn't handle certain edge cases, such as positive tab indexes, and had to deal with the focus unexpectedly escape the trap.
But once there's support for positive tab indexes (only for single container traps) there's no way I can think of to have the focus unexpectedly escape the trap (assuming the user doesn't create a multi-container trap with positive tab indexes).

It should be noted, though, that removing `focusin` would change the behaviour of the library in regard to multi-container traps with positive tab indexes, since in that case the focus could escape and there would be no `focusin` to bring it back.

You may want to leave `findNextNavNode` as it is and create a new handler just for `keydown`, then always use the new handler unless the user has a multi-container trap with positive tab indexes, in that case fallback to use the legacy logic that doesn't properly handle positive tab indexes, but still doesn't let the focus escape.

But in my humble opinion, it's unnecessary to keep support for a case that is not supported anymore (positive-tab indexes in multi-container traps).
It would be acceptable to just document that the behaviour for that edge case changed since version x.x.x

Shall we remove `focusin`?

DaviDevMod: Sorry, even though there's no way for the focus to escape the trap via keyboard navigation, there's still the possibility that focus doesn't enter the trap to begin with.

After the user focuses an element that doesn't belong to any container and tabs away from it, the focus may be given to an element outside of the trap.
Handling this case on `keydown` requires comparing the document position of `event.target` against the containers, which is not trivial because of the shadow DOM.

DaviDevMod: Sorry, even though there's no way for the focus to escape the trap via keyboard navigation, there's still the possibility that the focus doesn't enter the trap to begin with.

After the user focuses an element that doesn't belong to any container and tabs away from it, the focus may be given to an element outside of the trap.

Handling this case on properly requires comparing the document position of `event.target` against the containers, which is not trivial because of the shadow DOM.

And while it could be handled improperly on `keydown` rather than on `focusin`, I guess that `focusin` is the preferred time because it doesn't prevent the browser from doing the right thing when possible.

index.js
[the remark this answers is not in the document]
stefcameron: These changes look good to me now, thank you!

index.js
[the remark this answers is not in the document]
stefcameron: This looks good now too, thank you!

index.js
stefcameron: I greatly appreciate your deep thinking here, but no, I do not support the idea of removing `focusin`. I'm 99.99999% sure that focus does not only escape due to positive tab indexes. The library has been dealing with focus escape for a very long time and I can just imagine all of what would potentially hit the fan if that handler wasn't there keeping an eye on things. 😅 

> And while the case could be handled improperly on `keydown` rather than on `focusin`, I guess that `focusin` is the preferred time because it doesn't prevent the browser from doing the right thing when possible.

I assume you meant to write "properly" instead of "improperly" here, but the second half of your sentence is the key: In as much as is possible, get out of the way and let the browser handle the focus. 🙂

stefcameron (approved):
@DaviDevMod Thank you for being willing to simplify your PR. I know it doesn't lead to perfect browser fidelity in those negative `tabindex` edge cases, but I think it gets close enough for an edge case that's seldom encountered (tabbing from a negative `tabindex` node at the edge of a container in a multi-container trap). I'm much more comfortable with these changes as they are now. 😄

DaviDevMod:
@stefcameron I agree that it is definitely not worth refactoring everything just to make this rare edge cases work, however you should consider refactoring everything just to make any kind of changes possible (and spot eventual bugs that at the moment are very hard to spot).
The code heavily relies on comments to be understood and even you are not 100% sure of how it works :laughing:

<details>
<summary>Working with it has been painful :sob:</summary>
<br>

For what I could see, an easy way to improve readability would be reducing the amount of nested conditions and concatenated `else if` by just [returning early](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) rather than keeping the return value in a variable and have [a single exit point](https://softwareengineering.stackexchange.com/a/118793). 
Eg, this one:

https://github.com/focus-trap/focus-trap/blob/e26c3aab541c825b403df5ca295f4b0e493c5041/index.js#L578

A more complex refactor, that may yield higher returns, would be one on the logic.
There are pieces of logic that are excessively complex because they are solving too many problems at the same time.
Like `findNextNavNode` handling both `keydown` and `focusin` events.
</details>

But I understand that refactoring the codebase, just because, may not be on top of your priorities and I really appreciate your dedication on maintaining this open source project helpful to thousands and thousands of people around the world (including myself).

I'm glad to see the issue finally closed :tada:

index.js
DaviDevMod: @stefcameron I actually meant to say "improperly".
If `target` doesn't belong to any container, the case can't be handled properly (because of the shadow DOM).
So it is handled improperly on `focusin`, which is better than handling it improperly on `keydown` :laughing:

stefcameron:
> The code heavily relies on comments to be understood and even you are not 100% sure of how it works 😆

Right! And those comments are only there because I added them over the course of the past 3 years since I took over as maintainer (it's the same for Tabbable and Focus-trap-react). They weren't there before, so they are simply the understanding I've come to based on experience so far. But there's a lot of history before this that I wasn't privy to, which is the hard part.

I totally agree about refactoring. In fact, I've long thought all 3 libraries are nearing the point of needing full re-writes with a fresh take on things and perhaps letting go of so much of the legacy API support. After all, it was created in IE6 days!! All the meaningful browsers of today are evergreen.

Another big ticket item is to consolidate the 3 libraries into a single monorepo to get rid of all the tooling duplication and maintenance work between all 3 repos, as well as inconsistent code patterns, repo configurations, and fix the fact that it's so difficult to debug these libraries in their current `budo` debug environments.

But, I'm not paid to do any of this, and I need to put bread on the table, so those things will have to wait a while longer.

Nonetheless, though not perfect, I do believe you and I still managed to move the needle forward here on a long-standing issue, in spite of the house of cards. 😉

index.js
stefcameron: Got it. 😉

PR #1253, opened 2024-09-04
feat: move Escape key handling to target phase for better `escapeDeactivates` control

### LINKED ISSUE #1247: Does `Escape` need to be handled in the capture phase?
Based on the [implementation notes](https://github.com/focus-trap/focus-trap/blob/master/index.js#L748-L767), I understand why `Tab` key presses need to be handled in the capture phase. However, does this same approach need to apply to `Escape`?

Currently, it seems impossible to ignore an `Escape` key event that should be handled and canceled (via `preventDefault()`) by a child of a focus-trapping component. For example, in an open, focused combobox inside a modal, `Escape` should close the combobox but not close the focus trap, which isn't possible due to event cancelation limitations in the child. Worth noting that these example components use the default target/bubbling phase. 

Do you think the event phase for `Escape` could be adjusted?

jcfranco:
Fixes #1247.

This change gives developers more flexibility within the `escapeDeactivates` callback as it will allow child elements to handle `Escape` key presses without causing the trap to close.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>


stefcameron:
Thank you! Just ACK'ing I saw this. I'll check it out in a few days.

stefcameron (approved):
Wow, _nailed it_! 🔨 If only all PRs could be as thorough as this.

stefcameron:
@all-contributors add @jcfranco for code

PR #1291, opened 2024-10-25
Fix fallbackFocus not used when initialFocus is selector to non-existing node

### LINKED ISSUE #1218: Fallback focus is not selected in the case of absence of initial focus element
Hi, 
I'm currently struggling with the situation that after creating a focus trap, it wants to focus on a certain element at the start (I use `initialFocus`). 
This element does not always exist. In this case, I wanted to use `fallbackFocus`. 
However, testing this case and reviewing the code I come to the conclusion that this does not work as I thought. 
In `getInitialFocusNode` the element to focus is retrieved by `let node = getNodeForOption('initialFocus')`. However, `getNodeForOption` in the absence of such an element throws an error `.... refers to no known node`. In my opinion, the `fallbackFocus` element should be focused in this scenario, but we don't get to that point in the code. Do I understand this behaviour correctly, or perhaps it works correctly?
I can create a PR if needed, or explain in more detail.

stefcameron:
Fixes #1218

If initialFocus is a string, which means a selector, and it points to a non-exitent node, per the docs, the fallbackFocus option should be used, but it wasn't.

Clarified the docs with more details on behavior between the two options, and fixed the bug.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>


stefcameron:
Need to test the manual demo tests __before merging__.

stefcameron:
Manual tests are verified. in-iframe demo doesn't work, but it looks like this is a Rollup config issue, not a code issue, as that demo is also already broken on the demo site today without this change.

PR #1340, opened 2025-01-08
Don't automatically unpause trap when paused manually

bernhardoj:
Currently, when the current focus trap is deactivated, the previous focus trap is automatically unpaused. In this PR, we want to fix the behavior so that:
1. If the previous trap is automatically paused, then automatically unpause it when the current trap is deactivated
2. If the previous trap is manually paused, then don't automatically unpause it when the current trap is deactivated 

<details>
<br/>

- Includes updated docs demo bundle if source/docs code was changed (run `npm run demo-bundle` in your branch and include the `/docs/demo-bundle.js` file that gets generated in your PR).  ✅
- E2E (i.e. demos) test coverage added/updated.  ✅
- Typings added/updated.
- Changes do not break SSR:
- Changeset added (run `npm run changeset` locally to add one, and follow the prompts).  ✅

</details>


cypress/e2e/focus-trap-demo.cy.js
```diff
@@ -1655,6 +1655,39 @@ describe('focus-trap', () => {
...
+        .findByRole('button', { name: /^deactivate trap 2/ })
+        .click({ force: true }); // click fails without `force` option
+
+      // focus can be transitioned freely when trap 2 is deactivated because trap 1 is still paused
+      cy.findByRole('heading', { name: 'focus-trap demo' })
+        .as('outsideEl')
+        .click();
+      verifyFocusIsNotTrapped(cy.get('@outsideEl'));
+    });
+  });
```
bernhardoj, line 1689: Added a test, but it always succeeds even without the fix. I noticed that when running the E2E test locally, the focus trap never works.

bernhardoj:
@stefcameron here is the PR to fix the auto unpause behavior

index.js
```diff
@@ -1016,41 +1020,63 @@ const createFocusTrap = function (elements, userOptions) {
...
+        const onPostUnpause = getOption(options, 'onPostUnpause');
+
+        onUnpause?.();
+
+        updateTabbableNodes();
+        addListeners();
+        updateObservedNodes();
+
+        onPostUnpause?.();
+      }
```
stefcameron, line 1049: Wrapping this in `if (paused !== state.paused) { ... }` would preserve the old behavior more closely in terms of code checks. Right now, on auto-pause, if already paused, it would still remove listeners (again) and update observed nodes (again).

Not sure if that ever happens (auto-pausing an already paused trap) but just to cover the "what if", we should only actually do the behavior if there's an actual change in the state.

index.js
```diff
@@ -1016,41 +1020,63 @@ const createFocusTrap = function (elements, userOptions) {
...
-
-      state.paused = true;
-      onPause?.();
+      state.manuallyPaused = true;
 
-      removeListeners();
-      updateObservedNodes();
+      if (state.paused) {
+        return this;
+      }
```
stefcameron, line 1063: Not necessary if the state check moves into `_setPausedState()` (see previous comment).

index.js
```diff
@@ -1016,41 +1020,63 @@ const createFocusTrap = function (elements, userOptions) {
...
-      const onPostUnpause = getOption(unpauseOptions, 'onPostUnpause');
-
-      state.paused = false;
-      onUnpause?.();
+      state.manuallyPaused = false;
 
-      updateTabbableNodes();
-      addListeners();
-      updateObservedNodes();
+      if (!state.paused || trapStack[trapStack.length - 1] !== this) {
```
stefcameron, line 1075: Can simplify to just `if (trapStack[trapStack.length - 1] !== this)` here with the state check added into `_setPausedState()`.

index.js
```diff
@@ -31,7 +31,10 @@ const activeFocusTraps = {
       trapStack.splice(trapIndex, 1);
     }
 
-    if (trapStack.length > 0) {
+    if (
+      trapStack.length > 0 &&
+      !trapStack[trapStack.length - 1]._isManuallyPaused()
+    ) {
       trapStack[trapStack.length - 1].unpause();
```
stefcameron, line 38: Given we don't have symmetry here by calling `_setPausedState(false)` (opposite of line 14), I think we need a comment here to mention why we're calling `unpause()` instead of `_setPausedState(false)` -- that being to ensure the `manuallyPaused` state flag is reset on auto-unpause.

Then again, I think we could safely call `_setPausedState(false)` here to make it symmetric because we've already ensured that `state.manuallyPaused` is `false` with the `&& !trapStack[trapStack.length - 1]._isManuallyPaused()` check. That's probably much better than the comment.

index.js
```diff
@@ -1016,41 +1020,63 @@ const createFocusTrap = function (elements, userOptions) {
       return this;
     },
 
+    _isManuallyPaused() {
+      return state.manuallyPaused;
+    },
+
+    _setPausedState(paused, options) {
```
stefcameron, line 1027: I was thinking we'd also make these non-enumerable on the trap object since we're wanting them to be internal methods.

```javascript
Object.defineProperty(trap, '_isManuallyPaused', {
  value() {
    return state.manuallyPaused;
  }
});

// and same for _setPausedState()
```

stefcameron (changes requested):
Thank you for making this new PR! This is pretty much bang on what I was thinking, just asking for a couple of tweaks. Nice work!

cypress/e2e/focus-trap-demo.cy.js
stefcameron: Hmmm.. That's a bit strange. I'll try to check it out locally myself.

stefcameron: If I remove the new check on line 36:
```
// !trapStack[trapStack.length - 1]._isManuallyPaused()
```
and try to run the e2e tests locally (`npm run test:e2e:chrome`), then the test fails with this error:
```
AssertionError: Timed out retrying after 4000ms: expected '<h1>' to be 'focused'
```
which I take to indicate that the final assertion, "focus can be transitioned freely when trap 2 is deactivated because trap 1 is still paused", fails because trap 1 got auto-unpaused since we're no longer checking for the manually-paused state.

bernhardoj: Oh, I ran the test with `npm run test:e2e:dev`. Using `npm run test:e2e:chrome` works.

index.js
bernhardoj: Done

index.js
bernhardoj: Done

index.js
bernhardoj: Done

index.js
bernhardoj: Done

index.js
bernhardoj: > Then again, I think we could safely call _setPausedState(false) here to make it symmetric because we've already ensured that state.manuallyPaused is false with the && !trapStack[trapStack.length - 1]._isManuallyPaused() check

Good point, so to confirm, we replace `trapStack[trapStack.length - 1].unpause();` with `_setPausedState(false)`, right?

stefcameron: > > Then again, I think we could safely call _setPausedState(false) here to make it symmetric because we've already ensured that state.manuallyPaused is false with the && !trapStack[trapStack.length - 1]._isManuallyPaused() check
> 
> Good point, so to confirm, we replace `trapStack[trapStack.length - 1].unpause();` with `_setPausedState(false)`, right?

Yes! More specifically, with `trapStack[trapStack.length - 1]._setPausedState(false);`

bernhardoj: Updated

stefcameron (approved):
👌

stefcameron (approved):

stefcameron:
I just added an extra sentence to the `changeset` to mention the bug you fixed at the same time (shouldn't be able to unpause a trap that isn't at the top of the stack; you can alter the new manually-paused state, but you can no longer actually get it to unpause).

Huge thanks again for doing this work. Saved me a lot of time when I needed time for higher priorities! 😄

stefcameron:
Will be published with #1342

bernhardoj:
Great, thanks for the review too!

bernhardoj:
Oh btw, should I open the PR on the `focus-trap-react` repo to update the `focus-trap` to 7.6.4 or will you handle the update?

Issue #1068, opened 2023-09-27
Bug with selecting jumping to first tabbable from last if web component is involved.


simonxabris:
## Description

I found an issue where if you put a focus trap on a web component, like you have a `my-modal` element, and also add `tabindex="0"` to it to make it focusable/tabbable, when you cycle through the tabbable elements within the web component with `Tab`, you cannot jump back to the first element from the last.

## Expected behaviour

When I hit `Tab` on the last element, the containing web-component, which is the first tabbable element is focused

## Actual behaviour

Hitting `Tab` on the last element seemingly does nothing

## Reproduction

More detailed steps on how you can see the issue are provided in this stackblitz

https://stackblitz.com/edit/vitejs-vite-mxdsxb?file=index.html

## Suggestion

I identified that the issue comes from here: https://github.com/focus-trap/focus-trap/blob/ed95b7caf463170f972e876bdb58c46c9264a122/index.js#L423-L425

The `node` that we want to focus is going to be the web-component itself, and `document.activeElement` will also be the web-component, but that's not really true, because inside its shadow-root, there's another activeElement, but that's not visible from the outside this way.

My proposed solution is that when doing this check, it should recursively check if `document.activeElement` has a shadow-root and if it does, check for the `activeElement` inside that.

If someone from the project can confirm that this is a bug and my solution seems okay, I'm happy to submit a PR. If you think another solution would work better, I'm happy to work on that as well.

stefcameron:
@simonxabris Thanks for bringing this to my attention, and for providing a repro sandbox. Very helpful.

It's interesting that it works in reverse, but not forward direction. It gets stuck going forward, but doesn't going backward.

The other thing I notice is that when the trap is first activated, the focusable/tabbable `<my-calendar>` slotted element doesn't show its focus rectangle, while I would expect it to since it's technically the first tabbable element inside the trap, which has the `<my-modal>` element as its root. I wonder why that is. Maybe it's focused but the focus rectangle just isn't showing, or maybe it's not focused (but it should be). Yet when going backward, it clearly gets focused and gets the rectangle around it.

In your investigation (which I also appreciate!), what has happening when pressing TAB with focus on "Day 3" (after which we'd expect it to go to `<my-calendar>`)?

I would expect we would be in `checkKeyNav()` and called `findNextNavNode()` -- does it return a `destinationNode`?

[...]

PR #1485, opened 2025-05-27
Allow activation element in shadow DOM to be auto-focused after trap is deactivated

marioliu-block:
Kind of related to https://github.com/focus-trap/focus-trap/pull/1072 / https://github.com/focus-trap/focus-trap/issues/1068

I noticed that support for shadow DOM focusing was added for custom elements within focus traps, but not for custom activation elements outside of focus traps.

This PR adds support for focusing on custom activation elements within the default deactivation behavior of the focus trap.

I tried writing Cypress tests for this change, but it doesn't seem to like custom elements and can't seem to focus on them.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -127,34 +127,6 @@
       </p>
     </div>
 
-    <div id="demo-aesd">
-      <h2 id="aesd-heading">activation element in shadow DOM</h2>
-      <p>
-        The activation button for this focus trap is inside an open <a href="https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM">shadow DOM</a>.
-      </p>
-      <p>
-        Deactivating the focus trap should put the focus back on the button within the shadow DOM.
-      </p>
-      <p>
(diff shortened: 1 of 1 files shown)


docs/index.html
```diff
@@ -127,6 +127,34 @@ <h2 id="animated-trigger-heading">animated trigger</h2>
       </p>
     </div>
 
+    <div id="demo-aesd">
```
stefcameron, line 130: Could you move this down to be right before the existing "in open shadow DOM" demo so that it's closer to that one and the "with shadow DOM" demo?

stefcameron (changes requested):
@marioliu-block Thank you for this great PR! You followed existing patterns really well which made reviewing it a breeze. 😄 I just have one tiny request if you don't mind, and then I'll happily merge this and get it published.

docs/index.html
marioliu-block: Done!

stefcameron (approved):

stefcameron:
@all-contributors add @marioliu-block for code, bug

stefcameron:
@marioliu-block Thank you for the update! Merging.

Issue #1575, opened 2025-08-19
Add built-in “screen-reader trap” (aria-hidden) so virtual cursor can’t escape the focus trap


stefcameron:
### Discussed in https://github.com/focus-trap/focus-trap/discussions/1557

<div type='discussions-op-text'>

<sup>Originally posted by **apenab** July 23, 2025</sup>
## 📝 Summary / Motivation
`focus-trap` keeps **keyboard** focus strictly inside a container, but most screen readers (VoiceOver, NVDA, JAWS) can still roam the rest of the page with their virtual cursor.  
Every consumer who builds a modal or drawer therefore adds custom code that:

1. Iterates over `document.body.children` (and sometimes Shadow DOM),
2. Applies `aria-hidden="true"` to everything outside the trap.

That leads to duplicated logic, inconsistent edge-case handling (nested modals, dynamic DOM) and increases the risk of accessibility regressions.

A first-class solution inside **focus-trap** would give developers “full trapping” out of the box—keyboard *and* assistive-tech navigation.

---

## 📋 Steps to Reproduce the Gap

1. Open the basic modal demo in any browser.  
2. Start a screen reader.  
3. Activate the modal → keyboard focus is trapped.  
4. Use SR browse/scan mode (VO rotor, NVDA arrows…).  
   You can still reach headings, links, etc. outside the modal.

---

## ✅ Expected Behaviour
While a trap is active, **everything** outside `trapContainer` should be:

* Undiscoverable by assistive technologies  
* Unfocusable and uninteractable (keyboard, mouse, touch, programmatic focus)

## 🙏 Thanks

Absolutely no rush—just floating the idea.
Happy to prototype a PR if maintainers think this belongs in core. Thanks for all your work on focus-trap! 🙌</div>

stefcameron:
Closing this. Seems there is no longer any interest. Can re-open later if that changes.

stefcameron:
We have a PR open to fix this now. Re-opening while it's under review.

stefcameron:
This solved with an `inert` strategy with #1691 Different from `aria-hidden`, but similar. Use the new `isolateSubtrees` option to enable it.

msev:
Thanks for the update @stefcameron .

It would be great to be able to choose between the `inert` attribute and the `aria-hidden="true"` attribute.

With the `aria-hidden="true"` attribute, we could maintain the proper functioning of the `clickOutsideDeactivates()` function.

[...]

PR #1691, opened 2025-12-05
Add config option to isolate trap subtrees

andy-blum:
This PR aims to solve issue #1575.

Right now, focus trap works perfectly for holding the browser's focus, but screen readers like VoiceOver and NVDA can still freely roam the document. This PR creates a new config option, `isolateSubtrees`, which when enabled starts from the trap's container(s) and assembles two sets:
* All container elements and their direct lineage toward the document root
* All the child elements of the elements in the first set

By then iterating over set one and removing those items from set two, we're left with a set of elements that are directly adjacent to our containers' lineages. We then iterate over this set and apply `inert`, preventing screen reader access.

<details>
<br/>

- Typings added/updated.
- Changes do not break SSR:

</details>

## Summary by CodeRabbit

* **New Features**
  * Added isolateSubtrees option (default: false) to optionally isolate/inert adjacent DOM when a trap is active.

* **Documentation**
  * Added README docs, a demo block, demo initialization, and UI controls showcasing subtree isolation.

* **Tests**
  * Added DOM fixture and test suite for isolateSubtrees scenarios; updated e2e selectors for nested-trap demo.

* **Chores**
  * Updated type definitions and release changeset.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


andy-blum:
@coderabbitai generate docstrings

index.js
```diff
@@ -983,6 +1043,9 @@ const createFocusTrap = function (elements, userOptions) {
       state.delayInitialFocusTimer = undefined;
 
       removeListeners();
+      if (config.isolateSubtree) {
+        setSubtreeIsolation(false);
```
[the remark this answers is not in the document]
andy-blum: After some researching, we don't have a memory leak that grows over time or is realistically in danger of consuming all available memory. We do have a memory leak issue in that we prevent garbage collection of references to nodes removed from the DOM. As it stands right now, if a reference to an element exists in a Set, and that element is removed from the DOM, the reference remains in the Set.

That being said, when the focus trap is made, we collect the references we'll need to isolate the subtree(s). We do this ahead of the trap's activation to reduce the compute time of the activation/unpausing. When a trap is paused or deactivated, we still want to hold those references in case the trap is going to be activated again. If the references need to be updated, the user can call `trap.updateContainerElements`, which will clear the Set entirely, and compile all new references.

The creation of the Set is a performance trade off between the (very small) potential for higher memory usage than needed and having to traverse the DOM every time we activate/unpause the trap which is likely to impact interaction-to-next-paint metrics. While a WeakSet may reduce the memory impacts, they are not iterable, and thus cannot be used to pre-collect the adjacent nodes to reduce INP times.

cypress/e2e/focus-trap-demo.cy.js
```diff
@@ -689,11 +689,11 @@ describe('focus-trap', () => {
 
       // activate outer trap and element in outer trap should be focused
       cy.get('@testRoot')
-        .findByRole('button', { name: /^activate trap/ })
+        .get('#activate-nested')
         .as('lastlyFocusedElBeforeTrapIsActivated')
         .click();
 
-      cy.findByRole('button', { name: /^deactivate outer trap/ })
+      cy.get('#deactivate-nested')
```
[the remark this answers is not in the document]
andy-blum: I updated the new demo to include a nested trap. When doing that, I made new buttons with identical text

docs/index.html
```diff
@@ -544,6 +544,93 @@ <h2 id="delay-heading-explicit">No delay</h2>
...
+            activate second, unrelated trap
+          </button>
+        </p>
+        <p>
+          <button id="deactivate-isolate-subtree" aria-describedby="isolate-subtree-heading">
+            deactivate outer trap
+          </button>
+        </p>
+
+        <!-- asdf -->
```
stefcameron: Should be removed?

docs/index.html
```diff
@@ -544,6 +544,93 @@ <h2 id="delay-heading-explicit">No delay</h2>
...
+          <p>
+            Here is a nested focus trap <a href="#">with</a> <a href="#">some</a> <a href="#">focusable</a> parts.
+          </p>
+          <p>
+            <button id="deactivate-nested-isolate-subtree">
+              deactivate nested trap
+            </button>
+          </p>
+        </div>
+        <!-- asdf -->
```
stefcameron: Comment should be removed?

index.js
```diff
@@ -982,6 +1098,10 @@ const createFocusTrap = function (elements, userOptions) {
       clearTimeout(state.delayInitialFocusTimer); // noop if undefined
       state.delayInitialFocusTimer = undefined;
 
+      if (config.isolateSubtrees) {
+        setSubtreeIsolation(false);
+      }
```
stefcameron: I think we could move these lines down to _just before_ calling `activeFocusTraps.deactivateTrap(...)` because that's the thing that's going to call `updateTabbableNodes()` and we need to un-inert everything before that happens. I don't think anything in the 4 statements immediately below (lines 1105-1108) care about the `inert` state.

And then whether we can move it down or not, could you add a comment before the `if` about this need to revert the `inert` state to its original state before deactivation causes a call to `updateTabbableNodes()`?

Future me is likely to forget why we stuck this here because the one thing I don't like about it (though I know it's necessary) is that on deactivation, we revert the `inert` state before even calling `onDeactivate()` which isn't symmetric with how we set the `inert` state just before `onPostActivate()` when activating a trap...

index.js
```diff
@@ -1074,11 +1202,16 @@ const createFocusTrap = function (elements, userOptions) {
...
         state.paused = paused;
+
         if (paused) {
           const onPause = getOption(options, 'onPause');
           const onPostPause = getOption(options, 'onPostPause');
           onPause?.();
 
+          if (config.isolateSubtrees) {
+            setSubtreeIsolation(false);
+          }
```
stefcameron: Might be necessary to not have this tied so closely to pausing/unpausing per my previous comment, and so we might want to remove this and replace it with more _precise_ subtree isolation calls...

index.js
```diff
@@ -1089,6 +1222,10 @@ const createFocusTrap = function (elements, userOptions) {
 
           onUnpause?.();
 
+          if (config.isolateSubtrees) {
+            setSubtreeIsolation(true);
+          }
```
stefcameron: Might be necessary to not have this tied so closely to pausing/unpausing per my previous comment, and so we might want to remove this and replace it with more _precise_ subtree isolation calls...

stefcameron:
@andy-blum Thanks for the latest update. __I think we're cooking now!__ But I found a bug while playing with your demo (which is super helpful, by the way, for manually testing this!)

isolate-subtrees-already-inert-bug

The problem is with the already-inert gray panel with the input and button.

When activating the outer trap and deactivating it, no problems. That panel remains inert.

But when activating the outer, and then either the nested or the 2nd trap (my recording only shows the outcome with the nested trap, but it's the same when activating the 2nd trap), and then deactivating the nested, and deactivating the outer, the gray panel is no longer inert.

I think something somewhere is doubling-up on un-inerting things, or deactivating the nested/2nd trap somehow is confusing the `alreadyInert` Set of the outer trap such that when the outer is deactivated, it maybe (I didn't have time to debug it tonight) somehow no longer has that panel in its `alreadyInert` Set. Strange.

That's the only bug I could find while playing with it. Should be clear for merging once that's fixed.

---

We're getting really close to Christmas now! 🎄 It's time for family and friends. Don't worry about trying to figure this out ASAP. I won't be looking at anything as of this Sunday for at least a week. It's totally fine to tie this off in the new year.

andy-blum:
@stefcameron I've found the issue and created a fix & a new test case for that. I'm having gemini stress test it and I'll push changes up shortly!

andy-blum:
Thanks for your reviews and patience as well! Looking forward to having this in our project!

stefcameron:
@all-contributors Add @andy-blum for code

stefcameron:
published as focus-trap@7.7.0

andy-blum:
Thank you!

PR #1886, opened 2026-05-15
fix: ensure parent trap resumes using innermost active element in shadow DOM

### LINKED ISSUE #1885: Shadow DOM: initial focus fallback uses host active element instead of innermost active element
### Description

When focus is inside shadow DOM, `getInitialFocusNode()` can fall back to using `document.activeElement` (the shadow host) instead of the actual focused element inside the shadow root.

This can lead to incorrect focus restoration/fallback behavior during trap activation/deactivation flows, especially with nested traps.

### Repro

https://codepen.io/jcfranco/pen/MYbjrGB (compare behavior with [light DOM-only version](https://codepen.io/jcfranco/pen/MYbjrGB?shadow=false)).

### Actual Behavior

Focus handling can treat the shadow host as the active element, which may cause focus to move to the host or follow an unintended fallback path.

### Expected Behavior

Initial focus handling should use the **innermost active element** by recursively traversing open shadow roots so focus behavior aligns with the user-visible focused element.

### Root Cause

In the [`getInitialFocusNode()` fallback branch, relying on `document.activeElement`](https://github.com/focus-trap/focus-trap/blob/master/index.js#L317-L318) is not sufficient in shadow DOM scenarios because it may resolve to the shadow host instead of the actual focused descendant.

jcfranco:
Fixes: #1885

## Summary

Fix nested shadow DOM trap handoff by resolving initial focus from the innermost active element.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of the currently focused element across nested Shadow DOM so focus traps resume and behave correctly with custom elements.

* **New Features**
  * Focus handling now tracks the innermost active element through nested shadow roots for more reliable focus restoration and re-focus guards.

* **Tests**
  * Added shadow DOM test suite and fixtures covering nested traps, custom elements, and complex focus interactions.

* **Chores**
  * Updated release metadata for a patch publish.

### DIFF
--- a/.changeset/wicked-windows-retire.md
+++ b/.changeset/wicked-windows-retire.md
@@ -1,0 +1,5 @@
+---
+'focus-trap': patch
+---
+
+Improve shadow DOM focus handling for nested traps.
--- a/.changeset/wicked-windows-retire.md
+++ b/.changeset/wicked-windows-retire.md
@@ -2,4 +2,4 @@
 'focus-trap': patch
 ---
 
-Improve shadow DOM focus handling for nested traps.
+Improve shadow DOM focus handling for nested traps by ensuring a parent trap resumes using the innermost active element ([#1885](https://github.com/focus-trap/focus-trap/issues/1885))
--- a/index.js
+++ b/index.js
@@ -307,7 +307,7 @@ const createFocusTrap = function (elements, userOptions) {
    *
    * @param {Document | ShadowRoot} el
    *
-  * @returns {HTMLElement|null} The element that currently has the focus. `null` if a focused element isn't found.
+   * @returns {HTMLElement|null} The element that currently has the focus. `null` if a focused element isn't found.
    **/
   const getActiveElement = function (el) {
     const activeElement = el.activeElement;
--- a/index.js
+++ b/index.js
@@ -313,7 +313,7 @@ const createFocusTrap = function (elements, userOptions) {
     const activeElement = el.activeElement;
 
     if (!activeElement) {
-      return;
+      return null;
     }
 
     if (
--- a/index.js
+++ b/index.js
@@ -301,6 +301,31 @@ const createFocusTrap = function (elements, userOptions) {
     return node;
   };
 
+  /**
+   * Gets the current activeElement. If it's a web-component and has open shadow-root
+   * it will recursively search inside shadow roots for the "true" activeElement.
+   *
+   * @param {Document | ShadowRoot} el
+   *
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -301,6 +301,31 @@ const createFocusTrap = function (elements, userOptions) {
     return node;
   };
 
+  /**
+   * Gets the current activeElement. If it's a web-component and has open shadow-root
+   * it will recursively search inside shadow roots for the "true" activeElement.
+   *
+   * @param {Document | ShadowRoot} el
+   *
+   * @returns {HTMLElement} The element that currently has the focus
```
stefcameron:
```suggestion
  * @returns {HTMLElement|null} The element that currently has the focus. `null` if a focused element isn't found.
```

My understanding of `activeElement` is it's either an `HTMLElement` or `null` if the focused element is in a doc tree that isn't a descendent of the current one.

index.js
```diff
@@ -301,6 +301,31 @@ const createFocusTrap = function (elements, userOptions) {
...
+   **/
+  const getActiveElement = function (el) {
+    const activeElement = el.activeElement;
+
+    if (!activeElement) {
+      return;
```
stefcameron: We should return `null` for consistency here.

.changeset/wicked-windows-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'focus-trap': patch
+---
+
+Improve shadow DOM focus handling for nested traps.
```
stefcameron:
```suggestion
Improve shadow DOM focus handling for nested traps by ensuring a parent trap resumes using the innermost active element ([#1885](https://github.com/focus-trap/focus-trap/issues/1885))
```

stefcameron (commented):
Thank you for this PR! Just a few minor loose ends.

index.js
jcfranco: This function and its doc already existed before my changes. I just needed to move it so it would be available to `getInitialFocusNode`.

I agree with your suggestions, but would you still like me to apply them here, or would you prefer a separate PR?

jcfranco: I'll make the changes, but LMK if you'd prefer a separate PR.

stefcameron: I realize it pre-existed your changes. It seemed like a good opportunity to clean it up a bit. Thanks for doing it!

stefcameron (approved):

stefcameron:
@all-contributors add @jcfranco for bug

